PR #5, opened 2022-10-31
Remove DD_AGENT_MAJOR_VERSION warning on versioned scripts

bkabrda:
We don't want to show the warning on the versioned scripts, as they represent explicit version choice.

NOTE: This also means that installing IoT Agent or Dogstatsd will no longer be possible with `install_script_agent6.sh` (unless `DD_AGENT_MAJOR_VERSION=7` is specified explicitly), but I think that's ok.


amenasria (approved):
LGTM !

PR #13, opened 2022-12-15
Add success and failure telemetry events

randomanderson:
This PR add success and failure telemetry events for agent installation. There are 3 main changes:

1.  The `report_telemetry()` function that sends telemetry to the backend
2. Calling `report_telemetry()` in every case there was an `exit`. This consists of:
     a. Setting `ERROR_MESSAGE`
     b. Setting `ERROR_CODE`
     c. Printing the message to the screen
     d. Reporting telemetry
3. Setting `report_failure_url` and `telemetry_url` was moved to earlier in the script to catch more potential errors.

### Testing
It was not possible to test every scenario, but I manually tested the following using the production telemetry intake:
* Successful install
* Successful upgrade
* Failure because of unknown minor revision (set via environment variable)
* Failure because of unknown flavor (set via environment variable)
* Failure because of unknown distribution channel (set via environment variable)
* Failure because of unsupported platform (mac)
* Setting `DD_INSTRUMENTATION_TELEMETRY_ENABLED=false` disables telemetry

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -960,8 +960,6 @@ if [ $no_start ]; then
   printf "\033[34m\n  * DD_INSTALL_ONLY environment variable set.\033[0m\n"
 fi
 
-# Metrics are submitted, echo some instructions and exit
-printf "\033[32m
 for current_service in "${services[@]}"; do
   nice_current_flavor=${flavor_to_readable[$current_service]}
 
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -127,6 +127,7 @@ END
     "$telemetry_url" \
     --header 'Content-Type: application/json' \
     --header "DD-Api-Key: $apikey" \
+    --header 'DD-Telemetry-Debug-Enabled: true' \
     --data @-
   then
     printf "Unable to send telemetry\n"
@@ -157,6 +158,11 @@ function get_email() {
 }
 
 function on_error() {
+    if [ -z "${ERROR_MESSAGE}" ] ; then
+      # Save the few lines of the log file for telemetry if the error message is blank
+      SAVED_ERROR_MESSAGE=$(tail -n 3 $logfile)
+    fi
+
     printf "\033[31m$ERROR_MESSAGE
 It looks like you hit an issue when trying to install the $nice_flavor.
 
@@ -164,6 +170,7 @@ Troubleshooting and basic usage information for the $nice_flavor are available a
 
     https://docs.datadoghq.com/agent/basic_agent_usage/\n\033[0m\n"
 
+    ERROR_MESSAGE=$SAVED_ERROR_MESSAGE
     ERROR_CODE=$GENERAL_ERROR_CODE
     report_telemetry
 
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -275,10 +275,9 @@ if [ -n "$DD_SITE" ]; then
     report_failure_url="https://api.${DD_SITE}/agent_stats/report_failure"
 fi
 
+telemetry_url="https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry"
 if [ -n "$DD_SITE" ]; then
     telemetry_url="https://instrumentation-telemetry-intake.${DD_SITE}/api/v2/apmtelemetry"
-else
-    telemetry_url="https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry"
 fi
 
 if [ -n "$TESTING_REPORT_URL" ]; then
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -32,6 +32,12 @@ RPM_GPG_KEYS=("DATADOG_RPM_KEY_CURRENT.public" "DATADOG_RPM_KEY_E09422B3.public"
 # DATADOG_RPM_KEY.public is only useful to install old (< 6.14) Agent packages.
 RPM_GPG_KEYS_A6=("DATADOG_RPM_KEY.public")
 
+# Error codes for telemetry
+GENERAL_ERROR_CODE=1
+UNSUPPORTED_PLATFORM_CODE=5
+INVALID_PARAMETERS_CODE=6
+UNABLE_TO_INSTALL_DEPENDENCY_CODE=7
+
 # Set up a named pipe for logging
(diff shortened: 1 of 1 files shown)


randomanderson:
@DataDog/agent-platform : I'm not able to request reviewers. Can one of you take a look at this?

install_script.sh.template
```diff
@@ -63,6 +63,68 @@ function report(){
...
+  fi
+
+   (cat <<END
+       $telemetry_event
+END
+       ) | curl --fail --silent --output /dev/null \
+    "$telemetry_url" \
+    --header 'Content-Type: application/json' \
+    --header "DD-Api-Key: $apikey" \
+    --header 'DD-Telemetry-Debug-Enabled: true' \
```
randomanderson, line 124: I will remove this header before merging

paullegranddc (approved):

install_script.sh.template
```diff
@@ -63,6 +63,68 @@ function report(){
...
+"
+  fi
+
+   (cat <<END
+       $telemetry_event
+END
+       ) | curl --fail --silent --output /dev/null \
+    "$telemetry_url" \
+    --header 'Content-Type: application/json' \
+    --header "DD-Api-Key: $apikey" \
```
KSerrania, line 123: How important is it for you that the `$apikey` here is the same as the one used by the Agent? 

The main reason I'm asking this is that the `DD_API_KEY` environment variable passed to the script (which becomes this `$apikey` variable) may not be the one that is set in the `datadog.yaml` file.

For instance, if a `datadog.yaml` config file already exists (due to a prior install), then the `DD_API_KEY` parameter of the script is ignored, and no changes to the `datadog.yaml` file are made, see this code snippet: https://github.com/DataDog/agent-linux-install-script/blob/0a6f442b2182f9a70cd8ed3ea2f50609bb3677c7/install_script.sh.template#L825-L842

In fact, we don't require the `DD_API_KEY` variable at all for upgrades, therefore the `$apikey` may be empty: https://github.com/DataDog/agent-linux-install-script/blob/0a6f442b2182f9a70cd8ed3ea2f50609bb3677c7/install_script.sh.template#L403-L409

install_script.sh.template
```diff
@@ -202,6 +270,22 @@ else
...
+report_failure_url="https://api.datadoghq.com/agent_stats/report_failure"
+if [ -n "$DD_SITE" ]; then
+    report_failure_url="https://api.${DD_SITE}/agent_stats/report_failure"
+fi
+
+if [ -n "$DD_SITE" ]; then
+    telemetry_url="https://instrumentation-telemetry-intake.${DD_SITE}/api/v2/apmtelemetry"
+else
+    telemetry_url="https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry"
+fi
```
KSerrania, line 282: Could you make these two sections follow the same format?
```suggestion
report_failure_url="https://api.datadoghq.com/agent_stats/report_failure"
if [ -n "$DD_SITE" ]; then
    report_failure_url="https://api.${DD_SITE}/agent_stats/report_failure"
fi

telemetry_url="https://instrumentation-telemetry-intake.datadoghq.com/api/v2/apmtelemetry"
if [ -n "$DD_SITE" ]; then
    telemetry_url="https://instrumentation-telemetry-intake.${DD_SITE}/api/v2/apmtelemetry"
fi
```

install_script.sh.template
```diff
@@ -94,6 +156,9 @@ Troubleshooting and basic usage information for the $nice_flavor are available a
 
     https://docs.datadoghq.com/agent/basic_agent_usage/\n\033[0m\n"
 
+    ERROR_CODE=1
```
KSerrania, line 159: Could the meaning of each error code be made explicit in a comment somewhere in the file? For instance, after a quick reading of the file, it seems like 6 is for input errors, 5 is for OS / packages not supported, 1 is for all other generic errors, etc. and it would be nice to have a recap of these in the file.

install_script.sh.template
```diff
@@ -680,6 +801,9 @@ else
 Please follow the instructions on the Agent setup page:
 
     https://app.datadoghq.com/account/settings#agent\033[0m\n"
+    ERROR_MESSAGE="Unsupported platform"
```
KSerrania, line 804: Why not use the above `printf` message as the `ERROR_MESSAGE` here, like you did in other places?

install_script.sh.template
yahya-mouman: Yes you are right they are already defined in this sheet here https://docs.google.com/spreadsheets/d/1GWk23Y5VRw0IK7wdhNMy10OZYTwODasI518deYiMM_Y/edit#gid=0

randomanderson: I assigned each code to a variable at the top of the file and used those through instead of numbers

install_script.sh.template
randomanderson: The api key is used to determine the org for the telemetry. If the api keys are different, I think in almost all cases, they'll belong to the same org.

In the extremely rare case they do belong to different orgs, I think we can live with that. The telemetry will be mostly used in aggregate across all orgs.

For upgrades, I think partial data is okay. I asked the rest of the team if we should just block telemetry on upgrades altogether instead of partial data.

install_script.sh.template
randomanderson: The message here is quite long and spans multiple lines. I guess I could have used just the first part of the message 🤷‍♂️

randomanderson: Actually, some of the other error messages are multiline as well. I updated to code to make it consistent

install_script.sh.template
randomanderson: The team agrees that we shouldn't worry about the update case

randomanderson:
@KSerrania I tested the scenarios again on production. Everything is looking good to merge once you approve.

One thing I noticed for Mac testing was that I no longer got "Unsupported Platform" because the script is invalid (mac doesn't support `declare` syntax used). To get around having blank telemetry, I grab the last couple lines of the log file if `ERROR_MESSAGE` is blank.

olivielpeau:
@KSerrania is out this week so tagging @bkabrda for a review

bkabrda (approved):
LGTM now, please don't forget to remove the `DD-Telemetry-Debug-Enabled: true` header before merging.

PR #15, opened 2022-12-23
[template] fips-proxy package updates

truthbk:
This PR adds a few changes to the script to better tackle the installation of the FIPS proxy. 

Specifically it improves a little bit the management of the FIPS proxy side of things, and also adds some config and service management steps to fully enable the services. 

It also addresses two minor issues:
- `dockertest.sh` script platform definition support
- updated buildimages tag

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -832,10 +832,11 @@ fi
 declare -a monitoring_services
 monitoring_services=( "datadog-agent" )
 
-printf "\033[34m\n  * DD_INSTALL_ONLY environment variable set.\033[0m\n"
+if [ $no_start ]; then
+  printf "\033[34m\n  * DD_INSTALL_ONLY environment variable set.\033[0m\n"
+fi
 
 for current_service in "${services[@]}"; do
-
   nice_current_flavor=${flavor_to_readable[$current_service]}
 
   # Use /usr/sbin/service by default.
@@ -872,7 +873,7 @@ for current_service in "${services[@]}"; do
   # Metrics are submitted, echo some instructions and exit
   printf "\033[32m  Your ${nice_current_flavor} is running and functioning properly.\n\033[0m"
 
-  if [[  " ${monitoring_services[*]} ~= " ${current_service} "  ]]; then
+  if [[ "${monitoring_services[*]}" =~ "${current_service}" ]]; then
     printf "\033[32m  It will continue to run in the background and submit metrics to Datadog.\n\033[0m"
   fi
   
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -237,18 +237,21 @@ flavor_to_system_service=(
     ["datadog-dogstatsd"]="datadog-dogstatsd"
 )
 system_service=${flavor_to_system_service[$agent_flavor]:-datadog-agent}
+system_fips_service=datadog-fips-proxy
 
 declare -A flavor_to_etcdir
 flavor_to_etcdir=(
     ["datadog-dogstatsd"]="/etc/datadog-dogstatsd"
 )
 etcdir=${flavor_to_etcdir[$agent_flavor]:-/etc/datadog-agent}
+etcdirfips=/etc/datadog-fips-proxy
 
 declare -A flavor_to_config
 flavor_to_config=(
     ["datadog-dogstatsd"]="$etcdir/dogstatsd.yaml"
 )
 config_file=${flavor_to_config[$agent_flavor]:-$etcdir/datadog.yaml}
+config_file_fips=$etcdirfips/datadog-fips-proxy.cfg
 
 agent_major_version=AGENT_MAJOR_VERSION_PLACEHOLDER
 # shellcheck disable=SC2050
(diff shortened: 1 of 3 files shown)


install_script.sh.template
```diff
@@ -795,50 +826,58 @@ fi
...
+    # Try to detect Upstart, this works most of the times but still a best effort
+    restart_cmd="$sudo_cmd stop $current_service || true ; sleep 2s ; $sudo_cmd start $current_service"
+    stop_instructions="$sudo_cmd stop $current_service"
+    start_instructions="$sudo_cmd start $current_service"
+  fi
 
-\033[0m"
+  if [ $no_start ]; then
+    printf "\033[34m\n${preamble}
+    The newly installed version of the ${flavor_to_readable[$current_service]} will not be started.
```
KSerrania, line 856: You can define a variable containing `${flavor_to_readable[$current_service]}` at the start of the loop, since it's reused multiple times.

install_script.sh.template
```diff
@@ -237,18 +238,21 @@ flavor_to_system_service=(
     ["datadog-dogstatsd"]="datadog-dogstatsd"
 )
 system_service=${flavor_to_system_service[$agent_flavor]:-datadog-agent}
+system_fips_service=datadog-fips-proxy
```
KSerrania, line 241: How about directly defining the `services` array here, given that it's only used in one place (when printing the end message)?

```suggestion
declare -a services
services=("${flavor_to_system_service[$agent_flavor]:-datadog-agent}")
if [ -n "$fips_mode" ]; then
  services+=("datadog-fips-proxy")
fi
```

install_script.sh.template
```diff
@@ -651,22 +658,24 @@ elif [ "$OS" = "SUSE" ]; then
...
+
   if [ -z "$sudo_cmd" ]; then
     ZYPP_RPM_DEBUG="${ZYPP_RPM_DEBUG:-0}" zypper --non-interactive install "${packages[@]}" ||:
   else
     $sudo_cmd ZYPP_RPM_DEBUG="${ZYPP_RPM_DEBUG:-0}" zypper --non-interactive install "${packages[@]}" ||:
   fi
 
   # If zypper fails to refresh a random repo, it installs the package, but then fails
-  # anyway. Therefore we let it do its thing and then see if curl was installed or not.
+  # anyway. Therefore we l-    declare -a packages
```
KSerrania, line 678: This seems to be a copy-paste error?

install_script.sh.template
```diff
@@ -795,50 +826,58 @@ fi
...
-
-    $start_instructions
+#
+declare -a services
+services=("$system_service")
+if [ -n "$fips_mode" ]; then
+  services+=("$system_fips_service")
+fi
+
+preamble="  * DD_INSTALL_ONLY environment variable set."
```
KSerrania, line 836: I find this section hard to read (it uses the `preamble` text to only show some lines for the first service, assuming that the second service is the FIPS proxy).

I would be more explicit, and make the conditionals compare the `current_service` variable with the `datadog-fips-proxy` string:
- do not show the preamble text if `$current_service` is `datadog-fips-proxy`
- do not show the "It will continue to run in the background and submit metrics to Datadog." line if `$current_service` is `datadog-fips-proxy`.

WDYT?

install_script.sh.template
```diff
@@ -795,50 +826,58 @@ fi
...
 
-\033[0m"
+  if [ $no_start ]; then
+    printf "\033[34m\n${preamble}
+    The newly installed version of the ${flavor_to_readable[$current_service]} will not be started.
+    You will have to do it manually using the following command:
+  
+    $start_instructions\033[0m\n"
+  
+    exit
```
KSerrania, line 861: I believe that `exit` will skip printing the messages for the `datadog-fips-proxy` service (if enabled), is that intended?

install_script.sh.template
```diff
@@ -795,50 +826,58 @@ fi
...
+  else
```
KSerrania, line 862: In any case, I believe that `else` is not needed, since the previous case will either `exit` the script, or `continue` into the next iteration.

install_script.sh.template
truthbk: Indeed, must've fumbled my fingers with vim shortcuts.

install_script.sh.template
truthbk: It was intended, but the more I think of it the less desirable it seems. I will change that to `continue`.

install_script.sh.template
truthbk: ACK

install_script.sh.template
truthbk: So the reason I set the preample up, is because I want it on the first iteration no matter what order we traverse the services in. If we add in the future a `datadog-vector` package that runs before, or we mess the services array order for whatever reason, the current logic would still behave correctly.

I'm willing to change it if you wish.

KSerrania: I understand. My doubt is on whether what we're targeting is really the "first service", because right now the `preamble` variable changes the behavior of two different things. If in the future we install `datadog-vector` + `datadog-agent` + `datadog-fips-proxy` (in this order), then it's less clear to me which services should have the `* DD_INSTALL_ONLY environment variable set.` line, and which services should have the `It will continue to run in the background and submit metrics to Datadog.` line, and the answer may not be the "first service" for both.

To me:
- the `* DD_INSTALL_ONLY environment variable set.` should stay on the first iteration, so maybe it should just be printed before the loop for simplicity (one additional `if [ $no_start ]; then` conditional seems more readable to me than carrying that `preamble` variable logic).
- the `It will continue to run in the background and submit metrics to Datadog.` line would still apply to the `datadog-agent` service (or all services that actively fetch & send metrics, in the future), no matter what its position is, so I would explicit target (or exclude) the services that should (not) print that message.

KSerrania (approved):

PR #18, opened 2023-01-04
Added system-probe installation using

guyarb:
Added an option to install system-probe (using default configuration) via environment variable `DD_SYSTEM_PROBE_ENSURE_CONFIG=true`

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -321,9 +321,9 @@ if [ -n "$DD_FIPS_MODE" ]; then
   fips_mode=$DD_FIPS_MODE
 fi
 
-system_probe_enabled=
-if [ -n "$DD_SYSTEM_PROBE_ENABLED" ]; then
-  system_probe_enabled=$DD_SYSTEM_PROBE_ENABLED
+system_probe_ensure_config=
+if [ -n "$DD_SYSTEM_PROBE_ENSURE_CONFIG" ]; then
+  system_probe_ensure_config=$DD_SYSTEM_PROBE_ENSURE_CONFIG
 fi
 
 agent_flavor="datadog-agent"
@@ -937,7 +937,7 @@ if [ -n "$fips_mode" ]; then
 fi
 
 # set the system-probe configuration
-if [ -n "$system_probe_enabled" ]; then
+if [ -n "$system_probe_ensure_config" ]; then
   if [ -e "$system_probe_config_file" ] && [ -z "$upgrade" ]; then
     printf "\033[34m\n* Keeping old $system_probe_config_file configuration file\n\033[0m\n"
   else
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -321,6 +321,11 @@ if [ -n "$DD_FIPS_MODE" ]; then
   fips_mode=$DD_FIPS_MODE
 fi
 
+system_probe_enabled=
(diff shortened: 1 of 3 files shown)


test/localtest.sh
```diff
@@ -79,4 +79,28 @@ if [ -n "${EXPECTED_TOOL_VERSION}" ]; then
     fi
 fi
 
+system_probe_config_file=/etc/datadog-agent/system-probe.yaml
+if [ -n "${DD_SYSTEM_PROBE_ENABLED}" ]; then
+    if [ -e "$system_probe_config_file" ]; then
+        echo "[OK] Found system-probe configuration file $system_probe_config_file"
```
val06, line 85: use here shell colors as well, green for OK, red for FAIL

val06 (changes requested):
cred

test/localtest.sh
guyarb: that's not being used in the entire script, I don't think it is in the scope of the PR

val06 (approved):

bkabrda:
So the variable is called `DD_SYSTEM_PROBE_ENABLED` but the service is not actually started/enabled. However, by default it can't really meaningfully start it because the example config is blank.

I think we should either still start/enable the service and inform the user that it might not actually run if the config file is empty; or we can rename the variable to be more descriptive and inform the user that it will just set up the config file. I can discuss with our team, but in the meanwhile I'd also like to think what your opinion is @guyarb. Thanks!

guyarb:
> So the variable is called `DD_SYSTEM_PROBE_ENABLED` but the service is not actually started/enabled. However, by default it can't really meaningfully start it because the example config is blank.
> 
> I think we should either still start/enable the service and inform the user that it might not actually run if the config file is empty; or we can rename the variable to be more descriptive and inform the user that it will just set up the config file. I can discuss with our team, but in the meanwhile I'd also like to think what your opinion is @guyarb. Thanks!

Hey @bkabrda 
first of all, I added USM to system-probe.yaml.examle in https://github.com/DataDog/datadog-agent/pull/14908/files, but when I ran the localtest today, and looked into system-probe.yaml.examle, i didn't see we have usm there. Did I miss anything in the process?

Regarding your point, even if I manually adding some content to `system-probe.yaml` and run `service datadog-agent start` I still do not see system probe running
What am I missing in the process?
We can add the log as you suggested

install_script.sh.template
```diff
@@ -806,6 +812,20 @@ if [ -n "$fips_mode" ]; then
...
+# set the system-probe configuration
+if [ -n "$system_probe_enabled" ]; then
+  if [ -e "$system_probe_config_file" ] && [ -z "$upgrade" ]; then
+    printf "\033[34m\n* Keeping old $system_probe_config_file configuration file\n\033[0m\n"
+  else
+    if [ ! -e "$system_probe_config_file" ]; then
+      $sudo_cmd cp "$system_probe_config_file.example" "$system_probe_config_file"
+    fi
+
+    $sudo_cmd chown dd-agent:dd-agent "$system_probe_config_file"
```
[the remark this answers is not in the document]
guyarb: made the change, thanks

guyarb:
@bkabrda bump

guyarb:
@bkabrda I addressed your comment, let me know if I'm missing anything

bkabrda (approved):

PR #22, opened 2023-02-15
Use `fips` option instead of hardcoded *dd_url when DD_FIPS_MODE is set

Kaderinho:
## Description

This PR does 2 things:

- It prevents from installing the FIPS proxy if the Agent minor version is lower than 41 or when the architecture is not x86_64
     - We do this because the `fips` option as been added in `7.41` and the FIPS proxy is not supported for versions older than 7.41
- It adds `fips` configuration into `datadog.yaml` instead of the bunch of `dd_url` options

## Questions

- I used `report_telemetry` to follow the pattern but I'm not really sure if this error is relevant for the instrumentation-telemetry team tbh (We don't except this error to be triggered a lot anyway)
-  Right now `ARM64` arch is not supported for FIPS proxy. The current error is 'no package was found' but should we add a more explicit message ? 
- `agent_minor_version_without_patch` value still contains patch information if `DD_AGENT_MINOR_VERSION` is set to an RC (e.g. `40.0~rc.1-1` becomes `40.0~rc`). I don't think it's important but it was triggering an error so I used `{agent_minor_version_without_patch%.*}`

#### Verify that version check is working
- Run `DD_API_KEY=<API_KEY> DD_SITE="datadoghq.com" DD_FIPS_MODE=1 DD_AGENT_MINOR_VERSION=40.0~rc.7-1 ./install_script.sh` (replace `./install_script` with the path to the tested script)
    - It should output `FIPS mode is only available since version 7.41.0 and requested minor version is 40.0~rc.7-1`

#### Verify that arch check is working

- On a non x86_64 host: run `DD_API_KEY=<API_KEY> DD_SITE="datadoghq.com" DD_FIPS_MODE=1 ./install_script.sh`
    - It should output `FIPS mode isn't available for your architecture`

#### Verify that `fips` option is set in `datadog.yaml`

- Run `DD_API_KEY=<API_KEY> DD_SITE="datadoghq.com" DD_FIPS_MODE=1 ./install_script.sh` (you might need to add `REPO_URL=datad0g.com DD_AGENT_DIST_CHANNEL=beta` if packages are not deployed on prod yet)
     - Installation should run smoothly
     - Verify with ps (e.g. `ps aux | grep datadog`) that the Agent and the FIPS proxy are running
     - Run `datadog-agent diagnose datadog-connectivity` and verify that there is no error (expect for `/api/v1/validate` if you're not using a GovCloud API Key) + that the connection is going through `localhost:9804` (look at the command output)
     - Open `/etc/datadog-agent/datadog.yaml` and verify that the top lines are:
```
# Configuration for the agent to use datadog-fips-proxy to communicate with Datadog via FIPS-compliant channel.

fips:
    enabled: true
    port_range_start: 9803
    https: false
```

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -496,6 +496,14 @@ if [[ "$agent_flavor" == "datadog-dogstatsd" ]]; then
     fi
 fi
 
+if [ -n "fips_mode" ] && { [[ -n "$agent_minor_version_without_patch" ]] && [[ "${agent_minor_version_without_patch%.*}" -lt 41 ]]; }; then
+    ERROR_MESSAGE="FIPS mode is only available since version AGENT_MAJOR_VERSION_PLACEHOLDER.41.0 and requested minor version is $agent_minor_version"
+    ERROR_CODE=$UNSUPPORTED_PLATFORM_CODE
+    printf "\033[31m$ERROR_MESSAGE\033[0m\n"
+    report_telemetry
+    exit 1;
+fi
+
 # Root user detection
 if [ "$(echo "$UID")" = "0" ]; then
     sudo_cmd=''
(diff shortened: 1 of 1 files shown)


sgnn7 (approved):

bkabrda (changes requested):

bkabrda:
:wave: Hey, thanks for the PR. It mostly LGTM, I just left one inline comment.

To answer your questions:
* Let's use `INVALID_PARAMETERS_CODE` as failure code as explained above.
* I think adding a more explicit error message on ARM64 would be great. With generic errors like "package not found" the users don't know if that is expected or not a lot of times.
* The `agent_minor_version_without_patch` doesn't support RC/beta versions ATM. You can keep the code you implemented for this and we'll figure out a systematic solution later on.

Kaderinho:
Thanks for the review @bkabrda 🙏 
I've addressed your comments here: 3d195cd71b219f2b8f782b31eafb5eb15d33895f

bkabrda (approved):
LGTM now :+1:

PR #26, opened 2023-02-24
Makefile targets to automate the release process

Pythyu:
### Description

To automate the install script release process we need two new targets that :
 - Updates the changelog and bumps the version (before the release)
 - Updates the changelog again and adds the .post suffix to the version (after the release)

### Changes : 
- Added target pre_release_\<new-version\> : 
     -  Updates install_script_version=\<new-version\> in install_script.sh.template. It removes the .post as well
     - Updates the CHANGELOG.rst that renames the Unreleased version on top to \<new-version\>
- Added target post_release_\<new-version\>:
    - Updates install_script_version=\<new-version\> in install_script.sh.template to install_script_version=\<new-version\>.post (adding the .post at the end).
    - Updates the CHANGELOG.rst by adding a new Unreleased "version" on top of the file.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -41,6 +41,6 @@ install_script_agent7.sh: install_script.sh.template
 
 pre_release_%:
 	$(eval NEW_VERSION=$(shell echo "$@" | sed -e 's|pre_release_||'))
-	sed -i -e "s|install_script_version=.*|install_script_version=${NEW_VERSION}|g" install_script.sh.template
-
+	sed -i "" -e "s|install_script_version=.*|install_script_version=${NEW_VERSION}|g" install_script.sh.template
+	sed -i "" -e "s|^Unreleased|${NEW_VERSION}|g" CHANGELOG.rst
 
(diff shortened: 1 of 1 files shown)


bkabrda (commented):
This looks good, but could we have couple improvements:

* Most of the time, we would just bump the minor version in `pre_release`_*, e.g. we would make `X.Y.Z.post` => `X.Y+1.Z`. Would it be possible (in addition to what you already have) also include a `pre_release_minor` that would automatically figure out the `Y+1` so I wouldn't have to type it?
* For the `post_release_*` target, we don't really need to know the version. This should just be `post_release` and it should just append `.post` to `install_script_version=.*`. As an additional preventive measure it could check that the `.post` suffix isn't already there and error out if it is (so it doesn't get accidentally run twice and create `.post.post`).

Pythyu:
Good points, implementing these improvements and I'm coming back to you.

bkabrda (approved):

PR #30, opened 2023-03-16
Fix pre-release handing

Pythyu:
## Description
+ Previous implementation was making impossible to install a pre-release version. This should fix it and still parse pre-release correctly.

+ Fix on error handling that didn't worked on certain/older shell : 
```
agent_minor_version=19.0 
if [[ "$agent_minor_version" -lt 35 ]]; then
  echo ERROR
else
  echo OK
fi
```
  On classic recent shell we get
```
ERROR
```
  but on more basic shell like `sh` we get 
```
sh: [[: 19.0: syntax error: invalid arithmetic operator (error token is ".0")
OK
```
## Testing instructions
Test the script with : 
  - A release version ( e.g : 7.43.0 )
  - A pre-release version ( e.g : 7.15.1~beta.1 )
  - An error case ( e.g : datadog-dogstatsd 7.19.0~rc.7 on aarch64. Expected output : ``The Datadog Dogstatsd is only available since version 7.35.0 for your architecture.``)

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -496,7 +496,7 @@ if [[ "$agent_flavor" == "datadog-dogstatsd" ]]; then
         report_telemetry
         exit 1;
     fi
-    if  [[ "$OS" == "Debian" ]] && [[ `uname -m` == "aarch64" ]] && { [[ -n "$clean_agent_minor_version" ]] && [[ "$clean_agent_minor_version" -lt 35 ]]; }; then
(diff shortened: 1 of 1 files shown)


KSerrania (approved):
Could you add testing instructions in your PR description? Eg. test the script with a release version, test with a pre-release version, test with a pre-release version that triggers one of the conditions using  `clean_agent_minor_version`.

Before releasing a new version of the install script, we will want to do some QA of the install script, and having these instructions will help.

KSerrania (approved):

PR #31, opened 2023-03-20
[Fix] Ubuntu 14 false positive tests fix + dpkg-query upstream version equivalent

Pythyu:
## Changes : 
  - Ubuntu 14 tests didn't trigger an error when ``EXPECTED_MINOR_VERSION != MINOR_VERSION``. Previously we used ``-ne`` in the condition. But in the case where ``MINOR_VERSION`` is empty, ``test $VERSION -ne $EMPTY_VERSION`` would result in an error returning false. Meaning that it wouldn't enter the if that was supposed to handle error and be considered as valid when it wasn't. That's why ``-ne`` has been replaced with ``!=`` that handle the comparaison as string so even empty string would trigger the error.
  - ``MINOR_VERSION`` is empty because ubuntu 14's dpkg-query version is 1.17.x which is inferior of 1.18.16 needed to use the format ``Upstream-Version``. Instead we manually parse the version manually with ``sed``

### DIFF
--- a/test/localtest.sh
+++ b/test/localtest.sh
@@ -33,8 +33,10 @@ echo -e "\n"
 MAJOR_VERSION=$(echo "$INSTALLED_VERSION" | cut -d "." -f 1)
 MINOR_VERSION=$(echo "$INSTALLED_VERSION" | cut -d "." -f 2)
 
+echo "${EXPECTED_MAJOR_VERSION}"
+echo "${MAJOR_VERSION}"
 
-if [ "${EXPECTED_MAJOR_VERSION}" -ne "${MAJOR_VERSION}" ]; then
+if [ "${EXPECTED_MAJOR_VERSION}" != "${MAJOR_VERSION}" ]; then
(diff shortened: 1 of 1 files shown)


test/localtest.sh
```diff
@@ -20,7 +20,7 @@ EXPECTED_MINOR_VERSION="${EXPECTED_MINOR_VERSION:-${DD_AGENT_MINOR_VERSION}}"
 if command -v dpkg > /dev/null; then
     apt-get install -y debsums
     debsums -c ${EXPECTED_FLAVOR}
-    INSTALLED_VERSION=$(dpkg-query -f='${source:Upstream-Version}' -W ${EXPECTED_FLAVOR})
+    INSTALLED_VERSION=$(dpkg-query -W ${EXPECTED_FLAVOR} | sed -E "s/[^\t]*\t//g" | sed -E "s/[0-9]+://g")
```
bkabrda, line 23: I think `cut -f2 | cut -d: -f2` would be more readable than the regexes. Would that work?

Pythyu: I'll check if it works the same 👍

Pythyu: It should work all the same and is way easier to read :D

PR #35, opened 2023-04-28
Use dnf's --best on all distros that have dnf

bkabrda:
In order to upgrade the Agent by rerunning the install script when dnf is used, we have to correctly detect it on all distros where it exists.


Pythyu (approved):
LGTM !

PR #39, opened 2023-05-12
Install injection libraries with agent install script

randomanderson:
(resubmission of https://github.com/DataDog/agent-linux-install-script/pull/36)

This PR adds the ability to install autoinstrumentation libraries after setting environment variables.

| Variable name | Description |
| ---------------|------------|
| DD_APM_INSTRUMENTATION_ENABLED | Install host and docker injection unless overridden with a more specific env variable |
| DD_APM_HOST_INJECTION_ENABLED |  Installs or prevents installation of host injection (via `LD_PRELOAD`) |
| DD_APM_DOCKER_INJECTION_ENABLED | Installs or prevents installation of docker injection ( via `daemon.json` ) |
| DD_APM_LIBRARIES | Space separated list of which languages to install support for. Valid values are `js`, `java`, `python`, `dotnet`, `all`.  Defaults to `all` if host or docker injection is enabled | 
| DD_NO_AGENT_INSTALL | If set, don't install the datadog agent. This is useful for the container injection only case. The check is for the existence of the variable, not any particular value. This mirrors DD_INSTALL_ONLY |

* The script does not verify that all combinations of new and existing variables make sense together (for instance `DD_NO_AGENT_INSTALL` and `DD_AGENT_MINOR_VERSION`)
* I removed the `Installing the $nice_flavor package` log statement because it is redundant with `Installing package(s) ...` line that follows and is misleading in some cases.

### Testing
I added few basic checks to the test suite. Only host injection could be tested because of the docker-in-docker issue.


bkabrda (approved):
LGTM and the CI is passing now. Thanks a lot for the PR, I'm merging it.

PR #43, opened 2023-05-25
Add commits in the changelog to help its generation

chouetz:
Add a list of commits from the previous tag to the changelog in the make targets.
I already filtered some commits that might be useless (related to releasing)
I'm not a big expert of Makefiles so my code looks dirty to me, don't hesitate to give feedback/recommendations

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,8 +40,7 @@ install_script_agent7.sh: install_script.sh.template
 	chmod +x $@
 
 pre_release_%:
-	$(eval CUR_VERSION=$(shell awk -F "=" '/^install_script_version=/{print $$NF}' install_script.sh.template))
-	$(eval CUR_VERSION=$(shell echo "${CUR_VERSION}" | sed -e 's|.post||'))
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -40,18 +40,32 @@ install_script_agent7.sh: install_script.sh.template
 	chmod +x $@
 
 pre_release_%:
+	$(eval CUR_VERSION=$(shell awk -F "=" '/^install_script_version=/{print $$NF}' install_script.sh.template))
+	$(eval CUR_VERSION=$(shell echo "${CUR_VERSION}" | sed -e 's|.post||'))
```
Pythyu, line 44: We can compact it into a one liner.
```suggestion
	$(eval CUR_VERSION=$(shell awk -F "=" '/^install_script_version=/{print $$NF}' install_script.sh.template | sed -e 's|.post||'))
```

Pythyu (changes requested):
LGTM !
Left a few nitpick.

Pythyu (approved):

PR #44, opened 2023-06-06
Add new APT and RPM signing keys

chouetz:
To prepare 2024 key rotation, we add the new signature keys for APT and RPM repositories


KevinFairise2 (approved):

bkabrda (approved):
LGTM, but when merging, please let's provide a more descriptive commit message that would make it clear what and how we're updating.

PR #45, opened 2023-06-07
Only replace top-level tags entry in config

bkabrda:
As of 7.45.0, there's an additional `# tags:` entry in the Agent config file, but that doesn't start at the beginning of the line. This PR ensures we don't accidentally replace that as well.


AlexandreYang (approved):

FlorentClarret (approved):

KevinFairise2 (approved):

KSerrania:
Tested on Ubuntu 22 and CentOS 6, worked on both.

PR #49, opened 2023-06-13
Fix APM variable names and behavior

jonbodner:
- DD_APM_INSTRUMENTATION_ENABLED has 3 valid values: host, docker, and all. 
  - all is undocumented. 
  - host turn on host injection
  - docker turns on docker injection
    - it also disables agent installation
- DD_APM_LIBRARIES is now called DD_APM_INSTRUMENTATION_LANGUAGES
- DD_NO_CONFIG_CHANGE can be set to prevent any datadog.yaml changes (use it when updating an existing agent, so we don't break a customer's config file. They will have to make the changes by hand)
- Add enabling remote config to the datadog.yaml changes

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -330,25 +330,45 @@ if [ -n "$DD_FIPS_MODE" ]; then
 fi
 
 declare -a apm_libraries
+declare host_injection_enabled
+declare docker_injection_enabled
 
-host_injection_enabled=
-if [[ "$DD_APM_HOST_INJECTION_ENABLED" == "true" || ( -z "$DD_APM_HOST_INJECTION_ENABLED" && "$DD_APM_INSTRUMENTATION_ENABLED" == "true") ]] ; then
-  if [ "$no_agent" ]; then
-      ERROR_MESSAGE="Host injection is not supported with the DD_NO_AGENT_INSTALL flag. Disable host injection with DD_APM_HOST_INJECTION_ENABLED=false or unset DD_NO_AGENT_INSTALL"
-      ERROR_CODE=$INVALID_PARAMETERS_CODE
-      echo -e "\033[31m$ERROR_MESSAGE\033[0m\n"
-      report_telemetry
-      exit 1;
-  fi
-
-  host_injection_enabled="true"
-  apm_libraries=("datadog-apm-inject")
+if [ -n "$DD_APM_INSTRUMENTATION_ENABLED" ]; then
+  read -r -a target_array <<< "$DD_APM_INSTRUMENTATION_ENABLED"
+  for target in "${target_array[@]}"
+  do
+    case $target in
+      all)
+        host_injection_enabled="true"
+        docker_injection_enabled="true"
+        ;;
+      docker)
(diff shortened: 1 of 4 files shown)


install_script.sh.template
```diff
@@ -330,25 +330,46 @@ if [ -n "$DD_FIPS_MODE" ]; then
...
+  for target in "${target_array[@]}"
+  do
+    case $target in
+      all)
+        host_injection_enabled="true"
+        docker_injection_enabled="true"
+        ;;
+      docker)
+        docker_injection_enabled="true"
+        no_agent=true
```
bkabrda, line 347: Does this mean that when I specify `host,docker`, I will end up with `no_agent=true`? I think that's not what we want, am I right?

install_script.sh.template
```diff
@@ -397,6 +418,11 @@ elif [ -n "$host_injection_enabled" ] || [ -n "$docker_injection_enabled" ]; the
   apm_libraries+=("datadog-apm-library-all")
 fi
 
+no_config_change=
+if [ -n "$DD_NO_CONFIG_CHANGE" ]; then
```
bkabrda, line 422: Can we rename this to `DD_APM_NO_CONFIG_CHANGE` or even `DD_APM_INSTRUMENTATION_NO_CONFIG_CHANGE`? This script manipulates multiple configuration files and I think `DD_NO_CONFIG_CHANGE` would be too broad and could be misunderstood.

bkabrda (changes requested):
Hi :wave: thanks for sending the PR. I think overall it looks good, but I left two comments inline to discuss.

install_script.sh.template
jonbodner: Good point, I've removed the support for multiple values. "all" will cover it.

install_script.sh.template
jonbodner: Renamed to `DD_APM_INSTRUMENTATION_NO_CONFIG_CHANGE`.

install_script.sh.template
```diff
@@ -330,25 +330,33 @@ if [ -n "$DD_FIPS_MODE" ]; then
...
-  apm_libraries=("datadog-apm-inject")
+  no_agent=true
+fi
+
+if [ -n "${host_injection_enabled}${docker_injection_enabled}" ] ; then
+    apm_libraries=("datadog-apm-inject")
+fi
+
+if [ -n "${host_injection_enabled}" ] && [ -n "${no_agent}" ] ; then
+    ERROR_MESSAGE="Host injection is not supported with the DD_NO_AGENT_INSTALL flag. Disable host injection with DD_APM_HOST_INJECTION_ENABLED=false or unset DD_NO_AGENT_INSTALL"
```
bkabrda, line 355: I didn't notice this on the first review run, but I think the message is incorrect. I think it should say `Disable host injection with DD_APM_INSTRUMENTATION_ENABLED=docker ...` instead of `DD_APM_HOST_INJECTION_ENABLED=false`, shouldn't it?

jonbodner: Text fixed.

PR #51, opened 2023-06-15
SLES 11 install script support

Pythyu:
- Replaced all hashmap with a if and switch statement to be bash 3 compatible
  - Added opensuse 13.2 tests that downgrade bash to bash 3 before executing ``localtest.sh``. This is done via the option ``--suse-old`` on ``dockertest.sh`` that sets ``old-suse-startup.sh`` as the new entrypoint of the image.
  
### Notes
  - A small concern about bash 3, The official SUSE 11 repository containing the bash 3 rpm no longer exist. So I found a mirror on internet, I'm not sure if it's ok to use it. The owner of the mirror is at ``https://ftp5.gwdg.de/`` they host a lot of mirror of open source projects. Since the repository of SUSE 11 when it was still up was opensource I think it should be fine but we'll need to check. (isn't blocking for this PR we can merge and update later to comply the best)


.gitlab-ci.yml
```diff
@@ -21,6 +21,23 @@ generate-scripts:
...
       - install_script_agent7.sh
 
+test-opensuse-13:
+  image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/docker_x64:$DATADOG_AGENT_BUILDIMAGES
+  tags: ["runner:docker"]
+  stage: test
+  dependencies: ["generate-scripts"]
+  parallel:
+    matrix:
+      - IMAGE: images/mirror/opensuse/archive:13.2
```
chouetz, line 31: As I did this modification on my PR, I suggest not to repeat `images/mirror/` for all images in a matrix and set this directly on line 37: `--image "registry.ddbuild.io/images/mirror/${IMAGE}`

.gitlab-ci.yml
```diff
@@ -21,6 +21,23 @@ generate-scripts:
...
+      - IMAGE: images/mirror/opensuse/archive:13.2
+        FLAVOR: datadog-dogstatsd
+      - IMAGE: images/mirror/opensuse/archive:13.2
+        FLAVOR: datadog-iot-agent
+  script:
+    - ./test/dockertest.sh --image "registry.ddbuild.io/${IMAGE}" --script install_script_agent6.sh --minor-version "${MINOR_VERSION}" --expected-minor-version "${EXPECTED_MINOR_VERSION}" --flavor "${FLAVOR}" --suse-old "true"
+    - ./test/dockertest.sh --image "registry.ddbuild.io/${IMAGE}" --script install_script_agent7.sh --minor-version "${MINOR_VERSION}" --expected-minor-version "${EXPECTED_MINOR_VERSION}" --flavor "${FLAVOR}" --suse-old "true"
```
chouetz, line 38: you could add the agent version as part of the matrix, wdyt?
```
parallel:
    matrix:
      - IMAGE: opensuse/archive:13.2
        FLAVOR: ['', datadog-dogstatsd, datadog-iot-agent]
        VERSION: [agent6, agent7]
script:
    - ./test/dockertest.sh --image "registry.ddbuild.io/images/mirror/${IMAGE}" --script install_script_"${VERSION}".sh --minor-version "${MINOR_VERSION}" --expected-minor-version "${EXPECTED_MINOR_VERSION}" --flavor "${FLAVOR}" --suse-old "true"
```
Not sure the empty flavour would work, you might have to keep the first line as is

install_script.sh.template
```diff
@@ -1144,7 +1173,7 @@ fi
 
 service_cmd="service"
 if [ "$SUSE11" == "yes" ]; then
-  service_cmd=`$sudo_cmd which service`
+  service_cmd=`$sudo_cmd which service || echo "service"`
```
chouetz, line 1176: It seems you are by-passing a possible error here (probably to deal with your old opensuse), could you add a comment to explain why?

test/dockertest.sh
```diff
@@ -31,6 +31,9 @@ while [[ $# -gt 0 ]]; do
     --no-agent)
       DD_NO_AGENT_INSTALL="$2"
       ;;
+    --suse-old)
+      DD_OLD_SUSE="$2"
```
chouetz, line 35: suse-old or OLD_SUSE?
I would prefer to have `old-suse`, wdyt?

install_script.sh.template
```diff
@@ -445,15 +445,44 @@ if [ -n "$DD_AGENT_FLAVOR" ]; then
...
+        DATA="datadog-dogstatsd"
+    elif [ "$1" = "flavor_to_etcdir" ] && [ "$2" = "datadog_dogstatsd" ]; then
+        DATA="/etc/datadog-dogstatsd"
+    elif [ "$1" = "flavor_to_config" ] && [ "$2" = "datadog_dogstatsd" ]; then
+        DATA="$etcdir/dogstatsd.yaml"
+    fi
+    printf '%s' "$DATA"
+}
+
+agent_flavor_formatted=$(echo "$agent_flavor" | tr "-" "_")
```
chouetz, line 482: Is there a reason for this change? Shell options convention is more with dashes (whereas variables are more with underscore), so I'm not sure we should change this

install_script.sh.template
```diff
@@ -445,15 +445,44 @@ if [ -n "$DD_AGENT_FLAVOR" ]; then
...
+                DATA="Datadog FIPS Proxy"
+                ;;
+             "datadog_heroku_agent")
+                DATA="Datadog Heroku Agent"
+                ;;
+              *)
+                DATA="Unknown"
+                ;;
+         esac
+    elif [ "$1" = "flavor_to_system_service" ] && [ "$2" = "datadog_dogstatsd" ]; then
```
chouetz, line 472: I'm not sure I would have set these additional maps (`flavor_to_system_service`, `flavor_to_etcdir` and `flavor_to_config`) in the same method: theses "maps" had only one data, and where used only once, so having the `if` block close to the usage would make more sense to me (rather than mixing this in a method with other purpose)
Then your generic "getMapData" would only focus on "flavor_to_readable", wdyt?

chouetz (commented):
Looks good to me, don't hesitate to give me your feedback on my suggestions

.gitlab-ci.yml
Pythyu: I could but since the install_script is not a big load for the runner I figured it would make more sense to not instantiate a a whole new instance only for a small difference. And I think since it's the way it's done in the script previously we could stay consistent with how it's done or edit all the other jobs as well but I think it's fine as it is. wdyt ?

install_script.sh.template
Pythyu: I think even with only one data inside of the Map, it was done this way for scalability purposes (otherwise we'd have just put a constant string value instead already). If I remove them outside of the generic "getMapData" then we'll loose in that regards. I think it's easy to maintain and still allow to add new flavor if needed wdyt ?

install_script.sh.template
Pythyu: Necessary no, just because otherwise we would have in ``getMapData`` Map names with underscore like ``some_map_name`` and flavor like ``some-flavor`` in the if else switch statement. So purely to improve readability if you think it's useless I'll remove it wdyt ?

install_script.sh.template
Pythyu: Yeah good idea 👍 will add a comment about that (TLDR; on the SLES 11 ``service`` exist but since we can't legally use an SLES 11 image because the system isn't free we're testing on a opensuse 13.2 with bash downgraded but marked as on SUSE11 we need to put "service" as the default value if the ``which`` fails like it will in SUSE 13.2)

test/dockertest.sh
Pythyu: good point 👍

.gitlab-ci.yml
chouetz: ok to keep like this, you are right it may save some compute (you can maybe just group the flavours and have the same behaviour as today)

install_script.sh.template
chouetz: by scalability you mean "possibility to add more values in the future"? If yes, I would prefer to stick with what we need (so yes constant values when there is no mapping to do), and introduce the map when we have several data to test.

install_script.sh.template
chouetz: indeed, because $1 is the old variable name and $2 the option name. I was just surprised because in your PR there is a change from `-` to `_` internally in the option name and I didn't see immediately this line, I guess it would be less confusing in the future. I'm ok with both, I let you chose :)

PR #52, opened 2023-06-20
Use dash for options

chouetz:
Noticed a usage where the mapping fails because the flavor also comes from hardcoded data inside the script (lines 489, 495, 507)
```shell
/usr/bin/systemctl
* Starting the Unknown...

  Your Unknown is running and functioning properly.
```
Restore usage of underscores in options as it would require change at all those places and would probably bring confusion


Pythyu (approved):
nice catch on the issue 👍 
LGTM !

PR #53, opened 2023-06-27
Prepare 1.18.0 ( and small fix in update_changelog )

KevinFairise2:


CHANGELOG.rst
```diff
@@ -2,9 +2,14 @@
 Release Notes
 =============
 
-Unreleased
+1.18.0
 ================
 
+- AP-2106 Add new GPG keys for APT and RPM repositories signature rotation (#44)
```
KSerrania, line 8: I'd remove the `AP-2106`, it's useful for internal tracking of PRs, but doesn't need to end up in the changelog.

CHANGELOG.rst
```diff
@@ -2,9 +2,14 @@
...
+- Restore usage of underscores in options as it would require change at several places and would probably bring confusion (#52)
+- SLES 11 install script support (#51)
```
KSerrania, line 10: Technically, SLES 11 was supported before, but we broke support in an earlier version, so I'd change this to:
```suggestion
- Fix install script on SLES 11 (#51, #52)
```

CHANGELOG.rst
```diff
@@ -2,9 +2,14 @@
 Release Notes
 =============
 
-Unreleased
+1.18.0
 ================
 
+- AP-2106 Add new GPG keys for APT and RPM repositories signature rotation (#44)
+- Restore usage of underscores in options as it would require change at several places and would probably bring confusion (#52)
```
KSerrania, line 9: This is a fix to another change in the same version (#51), and therefore doesn't need to be shown in the changelog.

KSerrania (approved):

PR #56, opened 2023-06-27
AP-2157 Modify CI to be able to trigger from datadog-agent

KevinFairise2:


.gitlab-ci.yml
```diff
@@ -21,84 +21,112 @@ generate-scripts:
...
+      --minor-version "${MINOR_VERSION}"
+      --expected-minor-version "${EXPECTED_MINOR_VERSION}"
+      --flavor "${FLAVOR}"
+      --old-suse "${OLD_SUSE}"
+      --apt-url "$APT_URL"
+      --apt-repo-version "$APT_REPO_VERSION_AGENT7"
+      --yum-url "$YUM_URL"
+      --yum-version-path "$YUM_VERSION_PATH_AGENT7"
+  variables:
+    OLD_SUSE: "false"
```
chouetz, line 49: question ❓ 
if you set it by default there are 2 definition of the variable in the `test-opensuse-13` case. Is there a risk the definition not to be done in the correct order (and resulting on OLD_SUSE=false in the test so not testing what we want silently)?

.gitlab-ci.yml
```diff
@@ -21,84 +21,112 @@ generate-scripts:
...
+    matrix:
+      - IMAGE: centos:centos7
+        MINOR_VERSION: 38
+      - IMAGE: rockylinux:9.0
+        MINOR_VERSION: 38
+      - IMAGE: ubuntu:14.04
+        MINOR_VERSION: 38
+      - IMAGE: debian:10.9
+        MINOR_VERSION: 38
+      # When installing pinned minor version and some repos are broken,
```
chouetz, line 82: I would prefer a jira ticket rather than commented code :)

.gitlab-ci.yml
```diff
@@ -21,84 +21,112 @@ generate-scripts:
...
         FLAVOR: datadog-dogstatsd
       - IMAGE: opensuse/archive:13.2
         FLAVOR: datadog-iot-agent
-  script:
-    - ./test/dockertest.sh --image "registry.ddbuild.io/images/mirror/${IMAGE}" --script install_script_agent6.sh --minor-version "${MINOR_VERSION}" --expected-minor-version "${EXPECTED_MINOR_VERSION}" --flavor "${FLAVOR}" --old-suse "true"
-    - ./test/dockertest.sh --image "registry.ddbuild.io/images/mirror/${IMAGE}" --script install_script_agent7.sh --minor-version "${MINOR_VERSION}" --expected-minor-version "${EXPECTED_MINOR_VERSION}" --flavor "${FLAVOR}" --old-suse "true"
+  variables:
+    OLD_SUSE: "true"
 
+test_pinned_version:
```
chouetz, line 65: question ❓ 
As you created this one, why not creating another one for `test_flavours`?
If all test start at same time having several matrixes or only one is basically the same so I guess your objective was only readability? Then we could set these appart too, wdyt?

chouetz (approved):
Only questions/suggestions, ok for me

.gitlab-ci.yml
KSerrania: This is not a new comment (this was just moved), the corresponding task is AP-1885.

.gitlab-ci.yml
KevinFairise2: I think it is safe, when you extend a job, extended job fields are overriden by the fields defined in the jobs with `extends:` attribute. So here all jobs extending `.test` job will have a variable `OLD_SUSE: false` unless we explicitly specify another value for this variable

.gitlab-ci.yml
KevinFairise2: The objective here was not really readbility but rather, being able to not trigger install-script test with pinned versions when the pipeline is triggered from another pipeline. 
Because when we trigger it from an external pipeline we want to run the test with the version corresponding to the pipeline that triggered it.
That's why we have 
```
  rules:
    - if: '$CI_PIPELINE_SOURCE == "push"'
 ```
In this job and not in the other with al the other test cases

.gitlab-ci.yml
chouetz: I don't necessarily comment only on new code, sorry for that

PR #57, opened 2023-06-28
[install_script] Unable to acquire the dpkg frontend lock when multiple instance of dpkg are running

Pythyu:
Fix an issue reported via a support card about the install script not working on ubuntu while another process is using dpkg.

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -770,18 +770,30 @@ elif [ "$OS" = "Debian" ]; then
     apt_trusted_d_keyring="/etc/apt/trusted.gpg.d/datadog-archive-keyring.gpg"
     apt_usr_share_keyring="/usr/share/keyrings/datadog-archive-keyring.gpg"
 
-    printf "\033[34m\n* Installing apt-transport-https, curl and gnupg\n\033[0m\n"
-    $sudo_cmd apt-get update || printf "\033[31m'apt-get update' failed, the script will not install the latest version of apt-transport-https.\033[0m\n"
-    # installing curl might trigger install of additional version of libssl; this will fail the installation process,
-    # see https://unix.stackexchange.com/q/146283 for reference - we use DEBIAN_FRONTEND=noninteractive to fix that
-    if [ -z "$sudo_cmd" ]; then
-        # if $sudo_cmd is empty, doing `$sudo_cmd X=Y command` fails with
-        # `X=Y: command not found`; therefore we don't prefix the command with
-        # $sudo_cmd at all in this case
-        DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg
-    else
-        $sudo_cmd DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg
-    fi
+    MAX_RETRY_NB=10
+    for i in $(seq 1 $MAX_RETRY_NB)
+    do
+        printf "\033[34m\n* Installing apt-transport-https, curl and gnupg\n\033[0m\n"
+        $sudo_cmd apt-get update || printf "\033[31m'apt-get update' failed, the script will not install the latest version of apt-transport-https.\033[0m\n"
+        # installing curl might trigger install of additional version of libssl; this will fail the installation process,
+        # see https://unix.stackexchange.com/q/146283 for reference - we use DEBIAN_FRONTEND=noninteractive to fix that
+        if [ -z "$sudo_cmd" ]; then
+            # if $sudo_cmd is empty, doing `$sudo_cmd X=Y command` fails with
+            # `X=Y: command not found`; therefore we don't prefix the command with
+            # $sudo_cmd at all in this case
+            DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg 2>/tmp/ddog_install_error_msg || cat /tmp/ddog_install_error_msg
+        else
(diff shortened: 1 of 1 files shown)


install_script.sh.template
```diff
@@ -770,18 +770,31 @@ elif [ "$OS" = "Debian" ]; then
...
+        if [ -z "$sudo_cmd" ]; then
+            # if $sudo_cmd is empty, doing `$sudo_cmd X=Y command` fails with
+            # `X=Y: command not found`; therefore we don't prefix the command with
+            # $sudo_cmd at all in this case
+            DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg 2>/tmp/ddog_install_error_msg || cat /tmp/ddog_install_error_msg
+        else
+            $sudo_cmd DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg 2>/tmp/ddog_install_error_msg || cat /tmp/ddog_install_error_msg
+        fi
+
+        if cat /tmp/ddog_install_error_msg | grep "Could not get lock"; then
```
chouetz, line 789:
```suggestion
        if grep "Could not get lock" /tmp/ddog_install_error_msg; then
```

install_script.sh.template
```diff
@@ -770,18 +770,31 @@ elif [ "$OS" = "Debian" ]; then
...
+            RETRY_TIME=$((i*5))
```
chouetz, line 790: why not a fixed time?

install_script.sh.template
```diff
@@ -770,18 +770,31 @@ elif [ "$OS" = "Debian" ]; then
...
+    do
+        printf "\033[34m\n* Installing apt-transport-https, curl and gnupg\n\033[0m\n"
+        $sudo_cmd apt-get update || printf "\033[31m'apt-get update' failed, the script will not install the latest version of apt-transport-https.\033[0m\n"
+        # installing curl might trigger install of additional version of libssl; this will fail the installation process,
+        # see https://unix.stackexchange.com/q/146283 for reference - we use DEBIAN_FRONTEND=noninteractive to fix that
+        if [ -z "$sudo_cmd" ]; then
+            # if $sudo_cmd is empty, doing `$sudo_cmd X=Y command` fails with
+            # `X=Y: command not found`; therefore we don't prefix the command with
+            # $sudo_cmd at all in this case
+            DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg 2>/tmp/ddog_install_error_msg || cat /tmp/ddog_install_error_msg
```
chouetz, line 784: will the `apt-get` return non 0 if we push something in stderr?
Do we really need to display the error log?

chouetz (changes requested):
Request changes for the suggestion I made, thanks in advance

install_script.sh.template
Pythyu: Waiting progressively allows to wait less if `apt-get update` is short and if it's very long you don't have to wake up every 3-5 sec to check

install_script.sh.template
Pythyu: We don't need it really but if your `apt-get install` fails it's easier to show the log than having to fetch the file manually after to understand what's happening. And yes it will return non 0 when there's an error to stderr.

chouetz (approved):
ok

install_script.sh.template
Pythyu: Yeah the only one AFAIK

install_script.sh.template
```diff
@@ -770,18 +770,31 @@ elif [ "$OS" = "Debian" ]; then
...
+            $sudo_cmd DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg 2>/tmp/ddog_install_error_msg || cat /tmp/ddog_install_error_msg
+        fi
+
+        if grep "Could not get lock" /tmp/ddog_install_error_msg; then
+            RETRY_TIME=$((i*5))
+            printf "\033[31mInstallation failed : Unable to get lock !!\nretrying in ${RETRY_TIME}s ($i/$MAX_RETRY_NB)\033[0m\n"
+            sleep $RETRY_TIME
+        else
+            break
+        fi
```
KSerrania, line 795: What happens if the `apt-get install -y apt-transport-https curl gnupg` command fails, but for other reasons? With this method, I think we are ignoring these kinds of errors.

install_script.sh.template
```diff
@@ -770,18 +770,31 @@ elif [ "$OS" = "Debian" ]; then
...
+            # `X=Y: command not found`; therefore we don't prefix the command with
+            # $sudo_cmd at all in this case
+            DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg 2>/tmp/ddog_install_error_msg || cat /tmp/ddog_install_error_msg
+        else
+            $sudo_cmd DEBIAN_FRONTEND=noninteractive apt-get install -y apt-transport-https curl gnupg 2>/tmp/ddog_install_error_msg || cat /tmp/ddog_install_error_msg
+        fi
+
+        if grep "Could not get lock" /tmp/ddog_install_error_msg; then
+            RETRY_TIME=$((i*5))
+            printf "\033[31mInstallation failed : Unable to get lock !!\nretrying in ${RETRY_TIME}s ($i/$MAX_RETRY_NB)\033[0m\n"
```
KSerrania, line 791: Let's not be too alarmist in log messages:
```
printf "\033[31mInstallation failed: Unable to get lock.\nrRetrying in ${RETRY_TIME}s ($i/$MAX_RETRY_NB).\033[0m\n"
```

PR #59, opened 2023-07-11
Makefile: fix in place sed on linux

chouquette:
sed -i "" gets rejected by GNU sed as it tries to interpret "" as a path


chouetz (changes requested):
I'm not sure of what will happen on macos with the current statement

chouetz (approved):

chouetz (approved):

PR #62, opened 2023-07-13
Prepare 1.19.0 release

chouquette:


chouetz (approved):

bkabrda (commented):
Let's make the change description more descriptive. E.g. we don't need the `[install_script]` string to prepend the first one. The entries should clearly describe what they do (e.g. `Update Changelog with APM variables PR` is not useful for users reading the changelog - it should say what was updated and how, same applies for the other two points).

chouquette:
> Let's make the change description more descriptive. E.g. we don't need the `[install_script]` string to prepend the first one. The entries should clearly describe what they do (e.g. `Update Changelog with APM variables PR` is not useful for users reading the changelog - it should say what was updated and how, same applies for the other two points).

I removed the changelog part and updated the message regarding dpkg retries, let me know if that works for you!

KSerrania (approved):

PR #63, opened 2023-07-14
Add initial OPW linux install script

spencergilbert:
~~**NOTE** this should be merged _after_ https://github.com/DataDog/dogweb/pull/100200 is released.~~ This has been merged.

Outstanding question on the versioning for this new script, and how we'd like to bump the agent installer version as well - will confirm with Agent Platform.

- Generate OPW script from the template
- Removed unused script functionality OPW doesn't use

OPW has a much smaller matrix of supported OS's and Arch's

- Only x86 and arm64
- Amazon Linux 2
- Debian 10, 11
- Ubuntu 20.04, 22.04
- RedHat 8, 9
- CentOS 8, 9

The bulk of the changes are removals from the original scripts, as well as renaming a number of terms for UX.

---

Discussed the install script with @bkabrda over Zoom, here are the main points we covered.

- We should remove the `report_telemetry()` function for now, but we understand the desire to include similar functionality in the future.
- Continue using `report()` regardless of the current "Agent" centric wording, [the controller](https://github.com/DataDog/dogweb/blob/prod/dogweb/controllers/agent_stats.py) will be updated to use more generic terms, and a field for `variant`, as well as having a new mailer so the OP team doesn't get Agent reports and vice versa.


install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+  if [ -n "$worker_minor_version" ] ; then
+    safe_worker_version=$(echo -n "$worker_major_version.$worker_minor_version" | tr '\n' ' ' | tr '"' '_')
+  else
+    safe_worker_version=$(echo -n "$worker_major_version" | tr '\n' ' ' | tr '"' '_')
+  fi
+
+  if [ -z "${ERROR_CODE}" ] ; then
+    telemetry_event="
+{
+   \"request_type\": \"apm-onboarding-event\",
```
sghall, line 83: Should this be something specific to OP?  This could be really helpful telemetry data 😄

spencergilbert: I actually have no idea where this goes, it sounds like a generic event that gets filtered down by the `event_name` and the tags - it would be good to know where this gets surfaced in the UI.

spencergilbert: I dug about the UI some and didn't find it, I expect we can get some more info next week from the platform team 👍

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+install_script_version="0.1.0"
+logfile="dd-install.log"
+support_email="support@datadoghq.com"
+
+# DATADOG_APT_KEY_CURRENT.public always contains key used to sign current
+# repodata and newly released packages
+# DATADOG_APT_KEY_382E94DE.public expires in 2022
+# DATADOG_APT_KEY_F14F620E.public expires in 2032
+# DATADOG_APT_KEY_C0962C7D.public expires in 2028
+APT_GPG_KEYS=("DATADOG_APT_KEY_CURRENT.public" "DATADOG_APT_KEY_C0962C7D.public" "DATADOG_APT_KEY_F14F620E.public" "DATADOG_APT_KEY_382E94DE.public")
```
bkabrda, line 20: OPW repos have never been signed with `DATADOG_APT_KEY_382E94DE.public`, so let's just omit it here.

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+
+# DATADOG_RPM_KEY_CURRENT.public always contains key used to sign current
+# repodata and newly released packages
+# DATADOG_RPM_KEY_E09422B3.public expires in 2022
+# DATADOG_RPM_KEY_FD4BF915.public expires in 2024
+# DATADOG_RPM_KEY_B01082D3.public expires in 2028
+RPM_GPG_KEYS=("DATADOG_RPM_KEY_CURRENT.public" "DATADOG_RPM_KEY_B01082D3.public" "DATADOG_RPM_KEY_FD4BF915.public" "DATADOG_RPM_KEY_E09422B3.public")
```
bkabrda, line 27: Same for `DATADOG_RPM_KEY_E09422B3.public` as above.

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+
+    gpgkeys="https://${keys_url}/DATADOG_RPM_KEY.public\n       https://${keys_url}/DATADOG_RPM_KEY_E09422B3.public"
+
+    gpgkeys=''
+    separator='\n       '
+    for key_path in "${RPM_GPG_KEYS[@]}"; do
+      gpgkeys="${gpgkeys:+"${gpgkeys}${separator}"}https://${keys_url}/${key_path}"
+    done
+
+    $sudo_cmd sh -c "echo -e '[datadog]\nname = Datadog, Inc.\nbaseurl = https://${yum_url}/${yum_version_path}/${ARCHI}/\nenabled=1\ngpgcheck=1\nrepo_gpgcheck=${rpm_repo_gpgcheck}\npriority=1\ngpgkey=${gpgkeys}' > /etc/yum.repos.d/datadog.repo"
```
bkabrda, line 416: The filename has to be something else than `datadog.repo`, e.g. `datadog-observability-pipelines-worker.repo` that the public OPW instructions propose. Same goes for the `[datadog]` repo ID - let's use what we have in https://docs.datadoghq.com/observability_pipelines/setup/datadog/?tab=rpmbasedlinux to be consistent.

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+      # is available, unless the --best flag is set
+      # NOTE: we assume that sometime in the future "/usr/bin/yum" will
+      # be removed altogether, so we test for that as well.
+      dnf_flag="--best"
+    fi
+
+    if [ -n "$worker_minor_version" ]; then
+        # Example: observability-pipelines-worker-1.2.1-1
+        pkg_pattern="$worker_major_version\.${worker_minor_version%.}(\.[[:digit:]]+){0,1}(-[[:digit:]])?"
+        agent_version_custom="$(yum -y --disablerepo=* --enablerepo=datadog list --showduplicates observability-pipelines-worker | sort -r | grep -E "$pkg_pattern" -om1)" || true
```
bkabrda, line 433: When we change the repo ID in the repofile, we'll have to change the `--enablerepo=<the-new-repo-id>`.

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+        verify_agent_version "-"
+    fi
+
+    declare -a packages
+    packages=("observability-pipelines-worker")
+    
+    echo -e "  \033[33mInstalling package(s): ${packages[*]}\n\033[0m"
+
+    $sudo_cmd yum -y --disablerepo='*' --enablerepo='datadog' install $dnf_flag "${packages[@]}" || $sudo_cmd yum -y install $dnf_flag "${packages[@]}"
```
bkabrda, line 442: Same as above, we'll need to change the `--enablerepo=...` argument.

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+        elif [ $apt_exit_code -ne 0 ]; then
+            cat $DD_APT_INSTALL_ERROR_MSG
+            exit $apt_exit_code
+        else
+            break
+        fi
+    done
+
+    printf "\033[34m\n* Installing APT package sources for Datadog\n\033[0m\n"
+    $sudo_cmd sh -c "echo 'deb [signed-by=${apt_usr_share_keyring}] https://${apt_url}/ ${apt_repo_version}' > /etc/apt/sources.list.d/datadog.list"
```
bkabrda, line 479: Similar what I said for RPM, let's keep consistent with the manual instructions in terms of the source list file name: https://docs.datadoghq.com/observability_pipelines/setup/datadog/?tab=aptbasedlinux

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+if [ ! -e "$etcdir"/pipeline.yaml ]; then
+  printf "\033[31mThe $nice_flavor won't start automatically at the end of the script because the pipeline configuration is missing,\nplease add one at $etcdir/pipeline.yaml and start the $nice_flavor manually.\n\033[0m\n"
+  no_start=true
+fi
+
+$sudo_cmd chown observability-pipelines-worker:observability-pipelines-worker "$config_file"
+$sudo_cmd chmod 640 "$config_file"
+
+# Creating or overriding the install information
+install_info_content="---
```
bkabrda, line 583: The install_info is something that only the Agent uses and submits it with the metadata payload, so we don't really need it here. OTOH, we will probably eventually need something like it for OPW as well, so I agree with keeping it.

bkabrda (changes requested):
Hi :wave: thanks for opening the PR. I think overall it looks good, but I left couple of suggestions inline to get addressed. I'll get back to you on the `report` and `report_telemetry` functions after I figure out what to do there.

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+    report_telemetry
+    exit 1;
+  fi
+  worker_dist_channel=$DD_WORKER_DIST_CHANNEL
+fi
+
+if [ -n "$TESTING_YUM_VERSION_PATH" ]; then
+  yum_version_path=$TESTING_YUM_VERSION_PATH
+else
+  yum_version_path="${worker_dist_channel}/observability-pipelines-worker-1"
```
bkabrda, line 342: The `-1` should be `-${worker_major_version}`. Same below for the `apt_repo_version`.

spencergilbert: Good call, lazy hardcoding here 😉

install_script_opw.sh
spencergilbert: I figured this would be an innocent enough inclusion for a unified installer, e.g. "just include all the keys any of the installers/packages could have used, an extra key or two shouldn't be _harmful_".

But yes, strictly unneeded here - will remove.

install_script_opw.sh
spencergilbert: Completely overlooked that, good catch!

install_script_opw.sh
spencergilbert:
```
sudo sh -c "echo 'deb [signed-by=/usr/share/keyrings/datadog-archive-keyring.gpg] https://apt.datadoghq.com/ stable observability-pipelines-worker-1' > /etc/apt/sources.list.d/datadog.list"
```

It looks like it matches the manual steps here? Should we update the manual steps as well to have a separate file name?

install_script_opw.sh
spencergilbert: Yep, it seemed like a "would be useful in the future, and easy enough to adjust here" so I opted to keep it in.

install_script_opw.sh
```diff
@@ -0,0 +1,644 @@
...
+
+https://docs.datadoghq.com/observability_pipelines/setup/"
+    ERROR_CODE=$UNSUPPORTED_PLATFORM_CODE
+    printf "\033[31m$ERROR_MESSAGE\033[0m\n"
+    report_telemetry
+    exit;
+fi
+
+# Set the configuration
+if [ "$apikey" ]; then
```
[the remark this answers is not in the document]
spencergilbert: I agree with the rule there, it gets a little tricky given how OPW has been packaged - when installed via either linux package the `bootstrap.yaml` file here will always be present - we don't have option to do the "copy the example into place if not present" logic which makes the "first time run" more difficult to manage.

I _guess_ we could install `yq` as a dependency, but I think I'd prefer to avoid that to keep the install footprint minimal.

What about including a machine readable comment indicating "this config file was written to by the install script, skip further changes"?

install_script_opw.sh
spencergilbert: @sghall - discussed with @bkabrda, and unfortunately this is unlikely to make the cut for the first version. We _do_ want to include it, and will make sure it gets back in as soon as the backend can handle the payloads properly.

install_script_opw.sh
spencergilbert: Resolved in [5bda0aa](https://github.com/DataDog/agent-linux-install-script/pull/63/commits/5bda0aa8b2b5d7b4c3262460d7012aff26b1a4c5)

install_script_opw.sh
spencergilbert: Resolved in [5bda0aa](https://github.com/DataDog/agent-linux-install-script/pull/63/commits/5bda0aa8b2b5d7b4c3262460d7012aff26b1a4c5)

install_script_opw.sh
spencergilbert: Resolved in [5bda0aa](https://github.com/DataDog/agent-linux-install-script/pull/63/commits/5bda0aa8b2b5d7b4c3262460d7012aff26b1a4c5)

install_script_opw.sh
spencergilbert: Resolved in [5bda0aa](https://github.com/DataDog/agent-linux-install-script/pull/63/commits/5bda0aa8b2b5d7b4c3262460d7012aff26b1a4c5)

install_script_opw.sh
spencergilbert: Resolved in [5bda0aa](https://github.com/DataDog/agent-linux-install-script/pull/63/commits/5bda0aa8b2b5d7b4c3262460d7012aff26b1a4c5)

install_script_opw.sh
spencergilbert: Resolved in [5bda0aa](https://github.com/DataDog/agent-linux-install-script/pull/63/commits/5bda0aa8b2b5d7b4c3262460d7012aff26b1a4c5)

install_script_opw.sh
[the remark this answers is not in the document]
spencergilbert: 🤔 adding a comment the first time doesn't really work either - we _might_ need to adjust the packaging itself to better model this guideline.

install_script_op_worker.sh
```diff
@@ -0,0 +1,553 @@
+#!/bin/bash
+# (C) Datadog, Inc. 2010-present
```
jszwedko, line 2: **💬 suggestion:** ‏ 

```suggestion
# (C) Datadog, Inc. 2023-present
```

install_script_op_worker.sh
```diff
@@ -0,0 +1,553 @@
...
+if [ -n "$TESTING_APT_URL" ]; then
+  apt_url=$TESTING_APT_URL
+else
+  apt_url="apt.${repository_url}"
+fi
+
+report_failure_url="https://api.datadoghq.com/agent_stats/report_failure"
+if [ -n "$DD_SITE" ]; then
+    report_failure_url="https://api.${DD_SITE}/agent_stats/report_failure"
+fi
```
jszwedko, line 208: **🔨 warning:** ‏ I'll look for @bkabrda's feedback here. It seems like we'll want to separate out install failures for the OPW.

install_script_op_worker.sh
```diff
@@ -0,0 +1,553 @@
...
+  ERROR_MESSAGE="Error starting ${nice_flavor}"
+
+  eval "$restart_cmd"
+
+  ERROR_MESSAGE=""
+    
+  # Metrics are submitted, echo some instructions and exit
+  printf "\033[32m  Your ${nice_flavor} is running and functioning properly.\n\033[0m"
+
+  printf "\033[32m  It will continue to run in the background and submit metrics to Datadog.\n\033[0m"
```
jszwedko, line 544: **🔨 warning:** ‏ This doesn't seem right :)

jszwedko (commented):
Overall this is looking good! It will be nice to bring this closer to the Agent script via a template to avoid needing to sync changes.

I am interested in seeing some tests here if we can.

install_script_op_worker.sh
spencergilbert: Ah, there was some feedback around this captured outside of the PR review - I believe the endpoint will remain the same for now, but the payload/handler adjusted here and in dogweb to account for the agent vs opw difference.

install_script_op_worker.sh
spencergilbert: _Technically_ it's true, OPW reports it's internal metrics and logs to Datadog - but I can drop this line if it's considered more wrong than right.

install_script_op_worker.sh
```diff
@@ -0,0 +1,553 @@
...
+    https://docs.datadoghq.com/observability_pipelines/\n\033[0m\n"
+
+    ERROR_MESSAGE=$SAVED_ERROR_MESSAGE
+
+    if ! tty -s; then
+      fallback_msg
+      exit 1;
+    fi
+
+    if [ "$site" == "ddog-gov.com" ]; then
```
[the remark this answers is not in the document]
spencergilbert: Would we want to fail that inside the install script, I can imagine that being forgotten if we ever do support Gov Cloud. It'd be easy enough to do that check and exit the script early if desired.

cc @barieom @austinchen15

install_script_op_worker.sh
spencergilbert: This was one of those "will probably be part of the template" so I opted not to change it.

spencergilbert:
> Overall this is looking good! It will be nice to bring this closer to the Agent script via a template to avoid needing to sync changes.

For the time being I believe the straight copy-paste was desired based on my conversation with @bkabrda. It doesn't feel super clean to build this from the template without more composability coming from [AP-2023](https://datadoghq.atlassian.net/browse/AP-2023).
 
> I am interested in seeing some tests here if we can.

👍 I have a partial implementation locally and should have that up by tomorrow.

install_script_op_worker.sh
```diff
@@ -0,0 +1,553 @@
...
+          * )
+            printf "Please answer yes or no.\n"
+            ;;
+        esac
+    done
+}
+trap on_error ERR
+
+# OPW doesn't have a public changelog
+function verify_agent_version(){
```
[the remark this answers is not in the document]
spencergilbert: Renamed in [40cd49d](https://github.com/DataDog/agent-linux-install-script/pull/63/commits/40cd49dd0a918e74c1b7a8c16d6ac5d51cad84ad)

install_script_op_worker.sh
jszwedko: Ah, gotcha 👍

install_script_op_worker.sh
jszwedko: Yeah, it's not a big deal. Just a bit confusing since it makes it seem like the script existed since 2010. Perhaps it'll end up being one of the templated variables.

test/op-worker-test.sh
```diff
@@ -0,0 +1,50 @@
...
+INSTALLED_VERSION=
+RESULT=0
+EXPECTED_MAJOR_VERSION=1
+EXPECTED_MINOR_VERSION="${EXPECTED_MINOR_VERSION:-${DD_OP_WORKER_MINOR_VERSION}}"
+
+if command -v dpkg > /dev/null; then
+  INSTALLED_VERSION=$(dpkg-query -W ${EXPECTED_FLAVOR} | cut -f2 | cut -d: -f2)
+else
+  INSTALLED_VERSION=$(rpm -q --queryformat "%{version}" "${EXPECTED_FLAVOR}")
+fi
```
spencergilbert, line 15: The `localtest.sh` also tests the installed package for changes, however since we're currently updating the `bootstrap.yaml` in place - this would always fail.

.gitlab-ci.yml
```diff
@@ -120,6 +121,22 @@ test-apm-injection:
...
+  stage: test
+  dependencies: ["generate-scripts"]
+  parallel:
+    matrix:
+      - IMAGE: images/mirror/amazonlinux:2
+      - IMAGE: images/mirror/debian:10
+      - IMAGE: images/mirror/debian:11
+      - IMAGE: images/mirror/rockylinux:9.0
+      - IMAGE: images/mirror/ubuntu:20.04
+      - IMAGE: images/mirror/ubuntu:22.04
```
spencergilbert, line 136: Pulled from https://datadoghq.atlassian.net/wiki/spaces/OPE/pages/2970288333/Observability+Pipelines+Worker+Supported+Platform+and+Versions missing a centos/rhel 8

spencergilbert (commented):
A solution to the edit-config-in-place problem would be to set all these values in `/etc/default/observability-pipelines-worker` which isn't shipped with the package, it doesn't seem as discoverable as the `bootstrap.yaml` file.

cc @austinchen15 as I used the config file based on my discussion with you.

install_script_op_worker.sh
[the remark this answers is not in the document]
austinchen15: I'm in favor of doing it in the script! I don't think it'll be forgotten if we do ever support Gov cloud. 

Besides, if we do, it will error out quickly and be a relatively easy error to fix as long as we leave a breadcrumb comment for future engineers to disable this logic upon Gov Cloud enablement.

install_script_op_worker.sh
spencergilbert: Added some additional context in the summary.

install_script_op_worker.sh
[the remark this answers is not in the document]
spencergilbert: Added.

spencergilbert:
Something started behaving _weirdly_ on this PR. I'm going to cleanup the git history on my branch and re-open.

PR #67, opened 2023-07-24
Minor template improvements

spencergilbert:
- Reduce duplication in CI config
- Use == for equality checks

Backporting changes suggested in https://github.com/DataDog/agent-linux-install-script/pull/66


chouetz (approved):
Thanks a lot!

PR #71, opened 2023-07-26
Add a new install_script_docker_injection script

bkabrda:
This PR adds a new script `install_script_docker_injection.sh` which sets `DD_APM_INSTRUMENTATION_ENABLED="docker"` in all cases.

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -87,6 +87,7 @@ function report_telemetry() {
     safe_agent_version=$(echo -n "$agent_major_version" | tr '\n' ' ' | tr '"' '_')
   fi
 
+  APM_TELEMETRY_SAFE_AGENT_VERSION_OVERRIDE_PLACEHOLDER
   if [ -z "${ERROR_CODE}" ] ; then
     telemetry_event="
 {
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -261,7 +261,7 @@ function ensure_config_file_exists() {
     fi
 }
 
-echo -e "\033[34m\n* Datadog AgentINSTALL_SCRIPT_REPORT_VERSION_PLACEHOLDER install script v${install_script_version}\n\033[0m"
+echo -e "\033[34m\n* Datadog INSTALL_SCRIPT_REPORT_VERSION_PLACEHOLDER install script v${install_script_version}\n\033[0m"
 
 hostname=
 if [ -n "$DD_HOSTNAME" ]; then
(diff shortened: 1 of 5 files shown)


install_script.sh.template
```diff
@@ -87,6 +87,7 @@ function report_telemetry() {
     safe_agent_version=$(echo -n "$agent_major_version" | tr '\n' ' ' | tr '"' '_')
   fi
 
+  APM_TELEMETRY_SAFE_AGENT_VERSION_OVERRIDE_PLACEHOLDER
```
KSerrania, line 90: IIUC, this is used to set the `agent_version` tag in the telemetry event to `noagent_autoinstrumentation` when using the `install_script_docker_injection.sh` script only. What happens if the `install_script_agent7.sh` script is used, with `DD_NO_AGENT_INSTALL` / `DD_APM_INSTRUMENTATION_ENABLED="docker"` set?

test/localtest.sh
```diff
@@ -6,14 +6,23 @@ if [ "${EXPECTED_FLAVOR}" != "datadog-agent" ] && echo "${SCRIPT}" | grep "agent
...
+fi
+/tmp/script.sh
 
 INSTALLED_VERSION=
 RESULT=0
 EXPECTED_MAJOR_VERSION=6
 if echo "${SCRIPT}" | grep "agent7.sh$" >/dev/null || [ "${EXPECTED_FLAVOR}" != "datadog-agent" ] ; then
     EXPECTED_MAJOR_VERSION=7
 fi
+if echo "${SCRIPT}" | grep "docker_injection.sh$" > /dev/null; then
```
KSerrania, line 23: Nice to have: we're repeating this conditional a few times, would be nice to set it once and reuse.

KSerrania (approved):
Left a few comments, but LGTM

test/localtest.sh
bkabrda: You're right, I'll fix that.

install_script.sh.template
bkabrda: With `install_script_agent7.sh` we'll just report the Agent version. This is intentional, because APM want to track how much the individual scripts get used and not so much how exactly they're used (at least for now - this may change in the future).

Even though `install_script_agent7.sh` will still support this case, all our docs will recommend using `install_script_docker_injection.sh` for this case.

KSerrania (approved):

PR #72, opened 2023-07-26
chmod the OPW envfile for better security

spencergilbert:


install_script_op_worker1.sh
```diff
@@ -505,6 +505,8 @@ if [ ! -e "$pipeline_file" ]; then
   fi
 fi
 
+$sudo_cmd chmod 640 "$env_file"
```
chouetz, line 508: Not sure if I'm paranoid or not, should you `chmod` before appending any secret to the file?

bkabrda: Agreed. Let's first create the file, then chmod and only then start writing values into it.

spencergilbert: Moved in https://github.com/DataDog/agent-linux-install-script/pull/72/commits/5b520df72356b3dddcd4219a0cef7cbb262c7fec

bkabrda (approved):

spencergilbert:
Force pushed as I rebased on `main` to pass CI.

PR #76, opened 2023-08-08
Kfairise/add test to install both agent

KevinFairise2:
Add more tests to try to install dd agent together with OPW agent

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -172,8 +172,21 @@ test-observability-pipelines-worker:
       - IMAGE: rockylinux:9.0
       - IMAGE: ubuntu:20.04
       - IMAGE: ubuntu:22.04
+      - IMAGE: amazonlinux:2
+      - IMAGE: amazonlinux:2023
+        INSTALL_CLASSIC_AGENT: --observability-pipelines-install-classic-agent install_script_agent7.sh
+      - IMAGE: debian:10
+        INSTALL_CLASSIC_AGENT: --observability-pipelines-install-classic-agent install_script_agent7.sh
+      - IMAGE: debian:11
+        INSTALL_CLASSIC_AGENT: --observability-pipelines-install-classic-agent install_script_agent7.sh
+      - IMAGE: rockylinux:9.0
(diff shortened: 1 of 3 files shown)


.gitlab-ci.yml
```diff
@@ -160,6 +160,8 @@ test-apm-injection:
 
 test-observability-pipelines-worker:
   image: 486234852809.dkr.ecr.us-east-1.amazonaws.com/ci/datadog-agent-buildimages/docker_x64:$DATADOG_AGENT_BUILDIMAGES
+  rules:
+  - if: '$CI_PIPELINE_SOURCE == "push"'
```
spencergilbert, line 164: **❓ question:** ‏ Was this for testing or intended to be left in? If intentional, why is it the only job with this rule?

.gitlab-ci.yml
```diff
@@ -172,8 +174,21 @@ test-observability-pipelines-worker:
       - IMAGE: rockylinux:9.0
       - IMAGE: ubuntu:20.04
       - IMAGE: ubuntu:22.04
+      - IMAGE: amazonlinux:2
```
spencergilbert, line 177: **💬 suggestion:** ‏ Missed the additional variable in this matrix variant.
```suggestion
      - IMAGE: amazonlinux:2
        INSTALL_CLASSIC_AGENT: --opw-install-classic-agent install_script_agent7.sh
```

test/op-worker-test.sh
```diff
@@ -1,7 +1,12 @@
 #!/bin/bash -e
 
+if [[ -v "$DD_OPW_INSTALL_CLASSIC_AGENT" ]]; then
+$DD_OPW_INSTALL_CLASSIC_AGENT
+fi
+
 $SCRIPT
 
+
```
spencergilbert, line 9: **🥜 nitpick:** ‏ Stray linebreak added?
```suggestion
```

spencergilbert (changes requested):

.gitlab-ci.yml
KevinFairise2: It was intended to be here, so this job is not triggered when this pipeline is triggered from `datadog-agent` repository.

spencergilbert (approved):

PR #77, opened 2023-08-21
Forward DD_ENV to datadog.yml [AP-2385]

chouquette:


install_script.sh.template
```diff
@@ -1122,6 +1127,10 @@ EOF
       formatted_host_tags="['""$( echo "$host_tags" | sed "s/,/','/g" )""']"  # format `env:prod,foo:bar` to yaml-compliant `['env:prod','foo:bar']`
       $sudo_cmd sh -c "sed -i \"s/^# tags:.*/tags: ""$formatted_host_tags""/\" $config_file"
   fi
+  if [ "$dd_env" ]; then
+    printf "\033[34m\n* Adding your DD_ENV to the $nice_flavor configuration: $config_file\n\033[0m\n"
+    $sudo_cmd sh -c "sed -i 's/# env:.*/env: $dd_env/' $config_file"
```
KSerrania, line 1132: There are multiple occurrences of `# env:` in the example config file ([here](https://github.com/DataDog/datadog-agent/blob/52538a086c55208de8031da4a9a0e85f1b3f5705/pkg/config/config_template.yaml#L154) and [here](https://github.com/DataDog/datadog-agent/blob/52538a086c55208de8031da4a9a0e85f1b3f5705/pkg/config/config_template.yaml#L1223)). The logic here should only modify the top-level `env` setting (like what's done for the `tags` config option).

chouquette: Indeed, thanks for noticing! This should now be fixed and accounted for in the tests

install_script.sh.template
```diff
@@ -1122,6 +1127,10 @@ EOF
       formatted_host_tags="['""$( echo "$host_tags" | sed "s/,/','/g" )""']"  # format `env:prod,foo:bar` to yaml-compliant `['env:prod','foo:bar']`
       $sudo_cmd sh -c "sed -i \"s/^# tags:.*/tags: ""$formatted_host_tags""/\" $config_file"
   fi
+  if [ "$dd_env" ]; then
```
chouetz, line 1130:
```suggestion
  if [ -n "$dd_env" ]; then
```
we might want to replace the other ones (as per [recommendation](https://google.github.io/styleguide/shellguide.html#testing-strings), to prevent confusion use explicit tests)

test/localtest.sh
```diff
@@ -103,6 +103,26 @@ if [ -n "${EXPECTED_TOOL_VERSION}" ] && [ -z "$DD_NO_AGENT_INSTALL" ]; then
     fi
 fi
 
+if [ "${EXPECTED_FLAVOR}" = "datadog-agent" ] && [ -z "$DD_NO_AGENT_INSTALL" ]; then
```
chouetz, line 106:
```suggestion
if [ "${EXPECTED_FLAVOR}" == "datadog-agent" ] && [ -z "$DD_NO_AGENT_INSTALL" ]; then
```

chouquette: I don't have a strong opinion on this, but this script is not POSIX compliant so we might as well use the bash niceties

install_script.sh.template
chouquette: Agreed, I'll take a look to update similar checks

install_script.sh.template
```diff
@@ -1122,6 +1127,10 @@ EOF
...
+    printf "\033[34m\n* Adding your DD_ENV to the $nice_flavor configuration: $config_file\n\033[0m\n"
+    $sudo_cmd sh -c "sed -i 's/^# env:.*/env: $dd_env/' $config_file"
```
chouetz, line 1132: not sure if we can do this
```suggestion
    $sudo_cmd sh -c "sed -i '0,/^# env/{s/^# env:.*/env: $dd_env/}' $config_file"
```
if we know `env` is supposed to be the first in config file it could help make the replacement more robust (be sure we don't do spare replacements)

chouquette: That seems sensible, but it would also imply that the yaml has duplicated keys and is therefor invalid no?

chouetz: yes it's risky, but for some replacements (e.g. tags) where we had issues, it could have  prevented errors

chouetz (approved):

PR #79, opened 2023-08-29
Minor cleanup in script template

chouquette:
This PR cleans up a few things in the script template:
- use `if [ -n "$FOO" ]` instead of `if [ foo ]`
- Remove an unneeded `$(echo foo)` to display `$foo`
- Use `$()` instead of backticks, as advised by shellcheck

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -506,7 +506,7 @@ fi
 
 declare -a services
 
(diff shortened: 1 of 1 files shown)


spencergilbert (approved):
Nice, I noticed these earlier as well 👍

PR #82, opened 2023-09-04
Prepare 1.21.0 release

chouquette:


CHANGELOG.rst
```diff
@@ -2,9 +2,16 @@
 Release Notes
 =============
 
-Unreleased
+1.21.0
 ================
 
+- Forward DD_ENV to datadog.yml [AP-2385] (#77)
```
Pythyu, line 8: Nitpick :
Not sure we want to keep reference to our internal jira with `[AP-2385]` in the changelog. I'd remove it

Pythyu (approved):
Left a comment otherwise LGTM !

CHANGELOG.rst
chouquette: Indeed that would make sense, I'll update the PR before tagging then. Thanks!

PR #83, opened 2023-09-20
Allow installation of arm64 FIPS Proxy packages

Kaderinho:
Arm was not supported for FIPS Proxy so a warning was added in #22.

We'll be releasing ARM packages in the next few weeks so we can include it in the install script.


amenasria (approved):
LGTM ! Banger !

PR #92, opened 2023-09-29
Show error summary when installation fails because of insufficient available disk space

Pythyu:
Grep specific part of the system's package manager error output to display them at the end to increase visibility when an installation fails because of insufficient available space.

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -172,6 +172,8 @@ function on_error() {
     printf "\033[31m$ERROR_MESSAGE
 It looks like you hit an issue when trying to install the $nice_flavor.
 
+    $ERR_SUMMARY
+
 Troubleshooting and basic usage information for the $nice_flavor are available at:
 
     https://docs.datadoghq.com/agent/basic_agent_usage/\n\033[0m\n"
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -169,6 +169,7 @@ function on_error() {
 
     printf "\033[31m$ERROR_MESSAGE
 It looks like you hit an issue when trying to install the $nice_flavor.
+    
     $ERR_SUMMARY
 
 Troubleshooting and basic usage information for the $nice_flavor are available at:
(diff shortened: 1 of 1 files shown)


install_script.sh.template
```diff
@@ -770,7 +772,9 @@ if [ "$OS" == "RedHat" ]; then
 
     echo -e "  \033[33mInstalling package(s): ${packages[*]}\n\033[0m"
 
-    $sudo_cmd yum -y --disablerepo='*' --enablerepo='datadog' install $dnf_flag "${packages[@]}" || $sudo_cmd yum -y install $dnf_flag "${packages[@]}"
+    $sudo_cmd yum -y --disablerepo='*' --enablerepo='datadog' install $dnf_flag "${packages[@]}" 2>"/tmp/ddog_install_error_msg" || $sudo_cmd yum -y install $dnf_flag "${packages[@]}" 2>"/tmp/ddog_install_error_msg"
+    cat "/tmp/ddog_install_error_msg"
```
chouquette, line 776: Is displaying the entire log expected or is this a left over debug?

chouquette (approved):
LGTM, modulo the question about `cat /tmp/ddog_install_error_msg`

install_script.sh.template
Pythyu: I kinda wanted to use something similar to `tee` to store stderr into a file and still show it on stderr this way it doesn't remove the error from the live logs. I didn't find a great alternative since `tee` only works for stdout and wasn't sure if redirecting stderr into stdout would be a great idea. At the moment I'm simply storing stderr and printing it back to avoid any information loss. But still not great because it messes up the order in the log if there's an error. If you have any suggestion I'm open to it 🤔

chouquette: Forcing errors onto `stdout` might be counter intuitive indeed, although currently we don't display anything and the errors are displayed on `stdout`, so as far as which FD we use, I guess that's fine
My worry was mostly about adding the entire log file to the terminal, since that wasn't the case before, and it might be a bit too verbose, but I'll let you decide as long as the change is intentional :)

install_script.sh.template
```diff
@@ -887,7 +891,9 @@ If the cause is unclear, please contact Datadog support.
 
     echo -e "  \033[33mInstalling package(s): ${packages[*]}\n\033[0m"
 
-    $sudo_cmd apt-get install -y --force-yes "${packages[@]}"
+    $sudo_cmd apt-get install -y --force-yes "${packages[@]}" 2>"/tmp/ddog_install_error_msg"
+    cat "/tmp/ddog_install_error_msg"
+    ERR_SUMMARY=$(grep "No space left on device" -A1 -B1 /tmp/ddog_install_error_msg || true)
```
chouetz, line 896:
```suggestion
    ERR_SUMMARY=$(grep -C1 "No space left on device" /tmp/ddog_install_error_msg || true)
```
I suppose it's equivalent?

install_script.sh.template
chouetz, line 776: Shouldn't we remove these `cat`. [I was told we prefer dogs (sorry for this one)]
The only display is the one in the final error message, correct?

chouetz (commented):
Do we do `cat` in addition to the final error message in case of errors?

chouetz (approved):

PR #98, opened 2023-10-06
[e2e] add e2e tests

pducolin:
Meant to run locally, for now

They implement in go https://github.com/DataDog/experimental/blob/main/teams/agent-platform/package-scripts-test/install_test.sh


test/e2e/agentplatform.go
```diff
@@ -0,0 +1,210 @@
...
+	if len(*af) > 0 {
+		return fmt.Errorf("flavor flag already set to %s while trying to set to %s", *af, value)
+	}
+	fmt.Printf("Setting flavor flag to %s\n", value)
+	*af = agentFlavor(value)
+	return nil
+}
+
+const (
+	scriptURL    = "https://dd-agent.s3.amazonaws.com/custom"
```
chouetz, line 35: It means we might state somewhere we need to push the install_scripts to be tested in the `custom` folder prior launching the tests, correct?

chouetz (commented):
Do we need a "install as much as possible" use case as well?
Regarding the install_script url, is it possible to pass a custom one (in case we don't necessarily put the script in `custom`?

chouetz (approved):

test/e2e/agentplatform.go
pducolin: True, will change it in a follow up PR to take a param or an env var

pducolin: Actually will move it in this PR to a flag

pducolin:
I'm closing this PR and opening a new, smaller one. Will add up based on it

PR #100, opened 2023-10-17
[e2e] part 2 refactoring

pducolin:
Refactoring previous PR that added e2e tests for the installer script.

In this iteration no feature is added, only refactoring code to move common code.

Best reviewed per commit to follow the refactoring.

Next iteration will add one file per mode, and will add two new modes: fips and maximal install, with different assertions - you can see the next iteration in https://github.com/DataDog/agent-linux-install-script/pull/98/files#diff-89059d4760ebc9a893c31db5ec0d36a1305d6e1d82ad43b4d8331e97b8e6552b

### DIFF
--- a/test/e2e/agentflavor.go
+++ b/test/e2e/agentflavor.go
@@ -1,0 +1,27 @@
+// Unless explicitly stated otherwise all files in this repository are licensed
+// under the Apache License Version 2.0.
+// This product includes software developed at Datadog (https://www.datadoghq.com/).
+// Copyright 2016-present Datadog, Inc.
+package e2e
+
+import "fmt"
+
+type agentFlavor string
+
+func (af *agentFlavor) String() string {
+	return string(*af)
+}
+
+func (af *agentFlavor) Set(value string) error {
+	if len(*af) > 0 {
+		return fmt.Errorf("flavor flag already set to %s while trying to set to %s", *af, value)
+	}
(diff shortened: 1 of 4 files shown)


pducolin:
For reviewers: job succeded on the CI: https://gitlab.ddbuild.io/DataDog/agent-linux-install-script/-/jobs/351058245

test/e2e/agentflavor.go
```diff
@@ -0,0 +1,27 @@
...
+	}
+	*af = agentFlavor(value)
+	return nil
+}
+
+const (
+	agentFlavorDatadogAgent     agentFlavor = "datadog-agent"
+	agentFlavorDatadogIOTAgent  agentFlavor = "datadog-iot-agent"
+	agentFlavorDatadogDogstatsd agentFlavor = "datadog-dogstatsd"
+)
```
spencergilbert, line 27: **❓ question:** ‏ Do we know what additional changes be needed to support `obersvability-pipelines-worker` or `vector`? It seems out of scope for these changes, but would be good to understand what future work may entail.

spencergilbert (approved):

test/e2e/agentflavor.go
pducolin: This test can potentially support any value supported by `DD_AGENT_FLAVOR`.

PR #101, opened 2023-10-17
[e2e] part 3: extract one test per mode

pducolin:
Remove need for mode flag


spencergilbert (approved):
Duplicating a bunch of the test is a bit of a bummer, but clear upgrade on the `else if` statement.

pducolin:
Do not ask why git found conflicts when there were not 🤷‍♀️ , maybe different commit hash

pducolin:
Passing e2e test: https://gitlab.ddbuild.io/DataDog/agent-linux-install-script/-/jobs/352128838

PR #107, opened 2023-10-26
add the ability to specify a version number for a tracing library.

jonbodner:
- add the ability to specify a version number for a tracing library using a : after the langauge name.
- remove datadog-apm-library-all, because it causes problems when you want to force an update to a latest version of a tracing library.

https://datadoghq.atlassian.net/browse/APMON-355


Kyle-Verhoog (commented):
probably worth splitting the changes for removing the all package and introducing the library versions since they're unrelated

jonbodner:
> probably worth splitting the changes for removing the all package and introducing the library versions since they're unrelated

fair point. I'll split it up.

install_script.sh.template
```diff
@@ -409,7 +409,48 @@ if [ -n "$docker_injection_enabled" ]; then
...
+        apm_libraries+=("datadog-apm-library-python${cur_version}")
+        ;;
+      dotnet)
+        apm_libraries+=("datadog-apm-library-dotnet${cur_version}")
+        ;;
+      ruby)
+        apm_libraries+=("datadog-apm-library-ruby${cur_version}")
+        ;;
+      *)
+        ERROR_MESSAGE="Unknown apm library: $cur_lib. Must be one of: all, java, js, python, dotnet, ruby"
```
bkabrda, line 445: The `all` shouldn't be here

install_script.sh.template
```diff
@@ -409,7 +409,48 @@ if [ -n "$docker_injection_enabled" ]; then
...
+    cur_version=$(echo "$lib" | cut -s -d':' -f2)
+    if [ -z "$cur_lib" ] ; then
+      cur_lib="$lib"
+    elif [ "$cur_version" = "latest" ] ; then
+      cur_version=""
+    else
+      cur_version=":$cur_version"
+    fi
+    case $cur_lib in
+      all)
```
bkabrda, line 426: We don't want this clause any more as we're removing "all"

bkabrda (approved):
LGTM :rocket:

bkabrda:
Actually one more comment - can we uninstall the `datadog-apm-library-all` explicitly as we want to move away from using it? It would be good to get it uninstalled from customer systems.

jonbodner:
Telemetry has been added since the original approval. Can we get a new approval now? I've check with @yshapiro-57 to make sure that the telemetry is being received correctly.

bkabrda:
I actually realized that there is one problem with this - if the installed versions are non-latest, the next `apt upgrade`/`dnf update` will upgrade them to latest versions. I guess this is not what we want here, right? If we want to avoid that, we would need to use locking mechanisms:

* https://manpages.ubuntu.com/manpages/focal/en/man8/dpkg-hold.8.html
* https://linux.die.net/man/5/yum.conf (search for `includepkgs`) - this would be a little more tricky, because it requires changing the actual repofile and listing all packages that are being installed, not just the tracer libraries.

Let me know if you need some help with this

jonbodner:
> I actually realized that there is one problem with this - if the installed versions are non-latest, the next `apt upgrade`/`dnf update` will upgrade them to latest versions. I guess this is not what we want here, right? If we want to avoid that, we would need to use locking mechanisms:
> 
> * https://manpages.ubuntu.com/manpages/focal/en/man8/dpkg-hold.8.html
> * https://linux.die.net/man/5/yum.conf (search for `includepkgs`) - this would be a little more tricky, because it requires changing the actual repofile and listing all packages that are being installed, not just the tracer libraries.
> 
> Let me know if you need some help with this

Hrm, that's a good point. We would also want to unlock when the script runs so that the script can perform the update.

How much work is it to get this added to the install script?

bkabrda:
It's not much work, I would say maybe 1 day including testing. For RPM, the repofile gets rewritten before the "yum" call is issued, so that will take care of updating the lock and things will work fine AFAICS. For DEB, you're right that we'll first need to unlock, then install and then lock again.

bkabrda:
@jonbodner I pushed a commit that implements the pinning - it'd be great if you could test it.

One caveat: with yum (CentOS/RHEL 6 and 7), if you pin a tracer and then try to rerun the script with a pin of lower version (so when you try to downgrade), the tracer will not be downgraded. This is a known yum behavior and would have to be worked around by uninstalling the tracer and then installing it. I think we can probably live with this, but I will leave it up to you to decide whether we want to invest more time in fixing this.

spencergilbert (approved):

install_script.sh.template
```diff
@@ -104,6 +104,39 @@ function report_telemetry() {
...
+          library_telemetry_info="${library_telemetry_info},\"specified_nodejs_lib_version\": \"${cur_version}\""
+          ;;
+        datadog-apm-library-python)
+          library_telemetry_info="${library_telemetry_info},\"specified_python_lib_version\": \"${cur_version}\""
+          ;;
+        datadog-apm-library-dotnet)
+          library_telemetry_info="${library_telemetry_info},\"specified_dotnet_lib_version\": \"${cur_version}\""
+          ;;
+        datadog-apm-library-ruby)
+          library_telemetry_info="${library_telemetry_info},\"specified_ruby_lib_version\": \"${cur_version}\""
```
chouquette, line 132: it feels like this block could be simplified by matching on the last segment on the string, something like this:
```
function get_apm_lib {
   case $1 in
  datadog-apm-library-*)
    case $1 in
      *-java|*-python|*-dotnet|*-ruby)
          echo ${1#datadog-apm-library}
          ;;
      *-js)
          echo nodejs
          ;;
      *)
         fail
    esac
     ;;
  *)
   fail
   ;;
esac
}

...

library_telemetry_info="${library_telemetry_info},\"specified_$(get_apm_lib $cur_lib)_lib_version\": \"${cur_version}\""
```

The same function could be reused in the similar block below

install_script.sh.template
```diff
@@ -432,7 +465,48 @@ if [ -n "$docker_injection_enabled" ]; then
...
+    if [ -z "$cur_lib" ] ; then
+      cur_lib="$lib"
+    elif [ "$cur_version" = "latest" ] ; then
+      cur_version=""
+    else
+      cur_version=":${cur_version}-1"
+    fi
+    case $cur_lib in
+      all)
+        apm_libraries+=("datadog-apm-library-java" "datadog-apm-library-js" "datadog-apm-library-python" "datadog-apm-library-dotnet" "datadog-apm-library-ruby" )
```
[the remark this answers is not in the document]
jonbodner: resolved

install_script.sh.template
```diff
@@ -466,7 +540,7 @@ if [ -n "$DD_APM_INSTRUMENTATION_LANGUAGES" ]; then
   done
 elif [ -n "$host_injection_enabled" ] || [ -n "$docker_injection_enabled" ]; then
   # Default to all libraries if injection is enabled
-  apm_libraries+=("datadog-apm-library-java" "datadog-apm-library-js" "datadog-apm-library-python" "datadog-apm-library-dotnet" "datadog-apm-library-ruby")
+  apm_libraries+=("datadog-apm-library-java" "datadog-apm-library-js" "datadog-apm-library-python" "datadog-apm-library-dotnet" "datadog-apm-library-ruby" )
```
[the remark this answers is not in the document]
jonbodner: resolved

PR #110, opened 2023-11-14
Generate an install signature on success and include in telemetry events

yshapiro-57:
## Description of changes

When we install the Datadog agent on any host, in particular, on a Linux host with the Linux install script, we want the install script to write an `install.json` file into the same directory that contains the Datadog agent config. The file needs to be in this format:

```
{"install_id":"45b492fb-385f-46d7-98d7-410db89f53e4","install_type":"linux_agent_installer","install_time":1699970223}
```

Here `install_id` is a UUID, `install_type` is the way the agent was installed on the host (for the Linux install script, it is the hardcoded string `linux_agent_installer`), and `install_time` is a Unix timestamp since the epoch of when we installed the agent.

If the `install.json` already exists at the time the install script runs, and is in the right format, we want to reuse the existing file instead of overwriting it with a new UUID and timestamp.

The reason we want to do this is because we want to measure the amount of time users are taking from installing the agent to sending the first trace. We will modify the agent (in other PRs like https://github.com/DataDog/datadog-agent/pull/20733, not in this PR) to read `install.json` and send its contents to the Datadog backend when the agent sends us telemetry.

In addition, the telemetry events that the agent installer sends today need to be modified to include the install_id, install_type and install_time as onboarding event tags, [as described here](https://github.com/DataDog/instrumentation-telemetry-api-docs/blob/main/GeneratedDocumentation/ApiDocs/v2/SchemaDocumentation/Schemas/tags.md#properties)

The tags only need to be send, and the `install.json` only needs to be generated, if the agent installation succeeds. If the installation fails, we don't want to leave an orphaned `install.json` behind.

## Motivation

Jira: [Update Agent Linux Installer to propagate install signature](https://datadoghq.atlassian.net/browse/AIT-8956)

RFC: [APM Onboarding time-to-value telemetry](https://docs.google.com/document/d/14vsrCbnAKnXmJAkacX9I6jKPGKmxsq0PKUb3dfiZpWE/edit#heading=h.8d3o7vtyu1y1)

## Testing

In addition to verifying that the CI is all green, I also tested [this part of the code](https://github.com/DataDog/agent-linux-install-script/blob/3f9beb7280830faaf759361fb61adf76e390ee86/install_script.sh.template#L1261-L1298) by copy-pasting it into a standalone Shell file, `install.sh`, and running it with `source ~/tmp/install.sh`. I had to make the following changes vs. the actual code:

- Add `etcdir=/opt/datadog-agent/etc` at the beginning to replicate what the agent does
- Add `sh -c "echo '$install_signature' > $etcdir/install.json"` and `echo $install_signature` at the end to observe what the script does
- I had to change single square brackets to double brackets [in this line](https://github.com/DataDog/agent-linux-install-script/blob/3f9beb7280830faaf759361fb61adf76e390ee86/install_script.sh.template#L1291) because of differences between bash and zsh (I was testing in zsh)

With this I verified that
- If `/opt/datadog-agent/etc/install.json` already exists and is in the right format, the script uses it as the install signature
- If the file exists but is in the wrong format, the script overwrites it with contents in the right format
- If the file does not exist, the script creates it
- If the file exists but is not writeable, the script generates its own new signature every time it runs, but it does not throw an error, meaning the installation will still succeed.

Test the installer by running `make; ./test/dockertest.sh --image ubuntu:22.04 --script install_script_agent7.sh` and verify that the install still succeeds:
```
[OK] Correct major version installed
[PASS] DD_AGENT_MINOR_VERSION not specified, not checking installed minor version
[OK] Correct tool_version found in install_info file
[OK] Expected environment was found
debsums: package datadog-apm-inject is not installed
[OK] datadog-apm-inject is not installed
```
## End to end testing
In your local code repository, change `datadoghq.com` to `datad0g.com` [in this line of code](https://github.com/DataDog/agent-linux-install-script/blob/3f9beb7280830faaf759361fb61adf76e390ee86/install_script.sh.template#L363).
Change `DD_API_KEY=123` to`DD_API_KEY=_YOUR_ACTUAL_API_KEY_FROM_STAGING_` in [this line of code](https://github.com/DataDog/agent-linux-install-script/blob/3f9beb7280830faaf759361fb61adf76e390ee86/test/dockertest.sh#L81). Use one of the API keys [from this page](https://dd.datad0g.com/organization-settings/api-keys) or create a new one if you don't have a staging API key already.

Then run a regular Docker test with `make; ./test/dockertest.sh --image ubuntu:22.04 --script install_script_agent7.sh`. Verify that this causes an `agent.installation.success` telemetry event to be written to the staging Postgres DB on our backend:
```
tracertelemetryintakedbus1staging=> select * from apm_telemetry.event_log where install_id is not null and event_name = 'agent.installation.success';
-[ RECORD 1 ]-------------+-------------------------------------
event_log_id              | 556bc4f4-c574-56f8-a051-76c36b97c923
service_config_id         | [null]
org_id                    | 2
event_name                | agent.installation.success
lib_language              |
lib_version               |
service_name              |
env                       |
agent_version             | 7
agent_platform            | native
product                   | [null]
connection                |
tracer_version_origin     |
script_version            | 1.23.0.post
user_email                | [null]
onboarding_attempt_window | 2023-11-15 00:00:00+00
error_message             |
error_code                | 0
time                      | 2023-11-15 00:20:35.551+00
install_id                | a5022999-c70a-4aa9-b8b5-f986261876aa
install_type              | linux_agent_installer
install_time              | 2023-11-15 00:20:33+00

Time: 15.622 ms
tracertelemetryintakedbus1staging=>
```


install_script.sh.template
```diff
@@ -1234,6 +1259,44 @@ if [ -n "$docker_injection_enabled" ]; then
...
+install_id=$(generate_install_id)
+install_type="linux_agent_installer"
+install_time=$(date +%s)
+
+# If an install.json already exists, is formatted correctly, and was generated by this script,
+# reuse the original install ID and time.
+if [ -f "$etcdir/install.json" ]; then
+  # Parse the JSON file using substring extraction to avoid a dependency on any JSON parser
+  install_info=$(cat "$etcdir/install.json" 2>/dev/null)
+  if [ ${#install_info} -eq 118 ]; then
```
spencergilbert, line 1290: **🥜 nitpick:** ‏ If we keep the existing `install_info` and new `install.json` files separate, the naming here is a little confusing. At first read I thought it was reading the existing file.

install_script.sh.template
```diff
@@ -74,13 +74,30 @@ function report(){
...
   if [ "$DD_INSTRUMENTATION_TELEMETRY_ENABLED" == "false" ] || \
     [ "$site" == "ddog-gov.com" ] || \
     [ -z "${apikey}" ] || \
     [ -z "$telemetry_url" ]; then
     return
   fi
 
+  install_id_tag=
+  install_type_tag=
+  install_time_tag=
```
spencergilbert, line 90: **❓ question:** ‏ These are instantiated empty because we're always injecting them into the tags?

spencergilbert (approved):

PR #111, opened 2023-11-17
remove datadog-apm-library-all

jonbodner:
Do not use datadog-apm-library-all to install every tracing library, because it doesn't allow us to easily force an upgrade to the latest version.

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -434,7 +434,7 @@ if [ -n "$DD_APM_INSTRUMENTATION_LANGUAGES" ]; then
   do
     case $lib in
       all)
-        apm_libraries+=("datadog-apm-library-all")
+        apm_libraries+=("datadog-apm-library-java" "datadog-apm-library-js" "datadog-apm-library-python" "datadog-apm-library-dotnet" "datadog-apm-library-ruby")
         ;;
       java)
         apm_libraries+=("datadog-apm-library-java")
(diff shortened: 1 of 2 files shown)


bkabrda:
@jonbodner I think we should keep library-all, because it's a good convenience wrapper for customers to "install all libraries". The solution to having install_script/ansible upgrade all of the tracing packages is actually independent of whether we have library-all or not.

In order to mitigate the "upgrade all currently installed tracers", we'll have to add specific commands to install_script/ansible, but we'll have to add these regardless of whether we have library-all or not. Hence I'd propose keeping library-all.

jonbodner:
> @jonbodner I think we should keep library-all, because it's a good convenience wrapper for customers to "install all libraries". The solution to having install_script/ansible upgrade all of the tracing packages is actually independent of whether we have library-all or not.
> 
> In order to mitigate the "upgrade all currently installed tracers", we'll have to add specific commands to install_script/ansible, but we'll have to add these regardless of whether we have library-all or not. Hence I'd propose keeping library-all.

If we want to keep library-all for the few people who will run apt/yum manually, that's fine. I don't think it should be used in our scripts since it makes it harder to encourage customers to use the latest versions of tracer libraries.

In practice, very few customers should be running library-all manually, since they should only install tracers for the languages that they are using. 

I also think we should be encouraging customers to only use specified versions of the tracer libraries, even with the install script and ansible. I have another branch with a fix for that that we'll get to after reInvent.

bkabrda:
Ok, I think I see your point, both `apt install datadog-apm-library-X`/`apt upgrade` and `dnf install datadog-apm-library-X`/`dnf update` will install latest versions of these libraries, so there is no need to have additional upgrade commands if we just do this. I think I see a reasonable path forward now how to deprecate this package for x86_64 as well, I'll communicate this offline (not directly related to this PR).

bkabrda (approved):

PR #113, opened 2023-11-21
CI: simplify test matrices

chouquette:
By using a more compacted form it will be easier to add more dimensions to those matrices later on.

More specifically, in order to migrate to k8s we'd ideally need to have a single job per major version as we won't be able to start containers from the job, which will require a new matrix entry.


chouetz (approved):
Nice one!

PR #50, opened 2023-06-15
Improve configuration update and unit test addition

chouetz:
Objective of this PR is to improve robustness of install_script.
The PR can be reviewed by commit:
- Simplification of method about file existence and ensure modification of configuration files only if they don't already exist
- Improve replacement methods for configuration update (to prevent multi matching)
- Add a linter in "e2e tests" to validate the generated configuration files are correct
- Extract all functions from install_script to be able to unit test them
- Split configuration update in several methods, and unit test them in a new UT job in test stage
---
Testing the script:
While I created separated functions, I added unit tests for all that concerns configuration + I moved the creation of configuration files to be sure they are not modified if they already exist.
Test performed:

**Multi config**
1. Remove any preinstalled agent and remove yaml files from /etc/datadog-agent
2. set several environment variable
```
export DD_API_KEY=4321
export DD_HOST_TAGS=foo:bar,baz:toto
export DD_REPO_URL=datad0g.com
export DD_SITE=monsite.com
export DD_HOSTNAME=monhote
export DD_RUNTIME_SECURITY_CONFIG_ENABLED=true
export DD_COMPLIANCE_CONFIG_ENABLED=true
```
3. launch install_script
4. Validate my values are in the various scripts and the logging information are correct

**Fips**
1. Remove any preinstalled agent and remove yaml files from /etc/datadog-agent
2. set several environment variable (same as above plus below one)
```
export DD_FIPS_MODE=true
```
3. launch install_script
4. Validate fips is set and neither site nor url are set 

**Existing files**
1. Uninstall the agent and keep yaml files (copy from examples)
2. Set the above env variables
3. launch install script
5. verify the values are not in yaml and logs are correct

PR #116, opened 2023-11-21
prevent file creation based on input variable

chouetz:
File creation behaviour was slightly changed by #50 
This PR reverts to the previous mechanism

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -1195,23 +1195,19 @@ function update_env(){
 }
 function update_runtime_security(){
   local sudo_cmd="$1"
-  local security_config="$2"
-  local local_config_file="$3"
-  if [ "$security_config" == "true" ]; then
-    printf "\033[34m\n* Enabling runtime security in $(basename "$local_config_file") configuration\n\033[0m\n"
-    $sudo_cmd sh -c "sed -i 's/^#\s*runtime_security_config:$/runtime_security_config:/' $local_config_file"
-    $sudo_cmd sh -c "sed -i '/^runtime_security_config:/,// s/\(\s\+\)#\s*enabled:\s*false/\1enabled: true/' $local_config_file"
-  fi
+  local local_config_file="$2"
+  ensure_config_file_exists "$sudo_cmd" "$local_config_file" "root"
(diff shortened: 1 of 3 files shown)


install_script.sh.template
```diff
@@ -268,11 +268,12 @@ function ensure_config_file_exists() {
     local sudo_cmd="$1"
     local config_file="$2"
     local owner="$3"
+    local valid_condition="$4"
 
     if [ -e "$config_file" ]; then
         printf "\033[34m\n* Keeping old $config_file configuration file\n\033[0m\n"
         return 1
-    else
+    elif [ -z "$valid_condition" ] || [ "$valid_condition" == "true" ]; then
```
KevinFairise2, line 276: **❓ question:** ‏
If `DD_RUNTIME_SECURITY_CONFIG_ENABLED` is not defined should we create the config file?

chouetz: you are right, I changed the condition, it should be better now

install_script.sh.template
```diff
@@ -1253,9 +1254,7 @@ fi
 
 # set the system-probe configuration
 if [ -n "$system_probe_ensure_config" ]; then
-  if ! ensure_config_file_exists "$sudo_cmd" "$system_probe_config_file" "root"; then
-    echo "System probe config file already created"
-  fi
+  ensure_config_file_exists "$sudo_cmd" "$system_probe_config_file" "root"
```
KevinFairise2, line 1257: I think that with this, the system probe config file will never be created since the `valid_condition` variable is not passed, right?

KevinFairise2: Maybe something like that could work
```
ensure_config_file_exists "$sudo_cmd" "$system_probe_config_file" "root" $system_probe_ensure_config
```
After deleting the if condition

chouetz: yes I need to change this now I changed the condition in the method, thanks a lot

chouetz: revert the changes on `ensure_config_file_exists` and changed the `update_runtime_security` and `update_compliance_configuration` methods
We should be back in the initial situation now, thanks for your help and sorry for inconvenience

KevinFairise2 (approved):

PR #118, opened 2023-11-22
dockertest: forward missing environment variable

chouquette:
The variable wasn't forwarded to the container so it wasn't actively used by the tests


chouquette:
I missed the part where it's used to dispatch to the appropriate entry point script, closing

PR #119, opened 2023-11-22
CI: pass parameters as environment variables

chouquette:
We will soon stop using dockertest.sh as part of the CI, but still need a way to provide our test parameters.

This PR provides those through environment variable that match what dockertest.sh does, and what the install script expect.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -155,16 +155,16 @@ test-apm-injection:
       - IMAGE: ["ubuntu:14.04", "ubuntu:22.04", "centos:centos7"]
         DD_APM_INSTRUMENTATION_ENABLED: "docker"
         NO_AGENT: "true"
-        APM_LIBRARIES: "all"
+        DD_APM_INSTRUMENTATION_LANGUAGES: "all"
         SCRIPT: "install_script_agent7.sh"
       - IMAGE: ["ubuntu:14.04", "ubuntu:22.04", "centos:centos7"]
         # We intentionally don't pass NO_AGENT here to ensure this script works without it
         # NO_AGENT: "true"
-        APM_LIBRARIES: "all"
+        DD_APM_INSTRUMENTATION_LANGUAGES: "all"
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -175,12 +173,14 @@ test-observability-pipelines-worker:
...
+  variables:
+    DD_OPW: 'true'
   parallel:
     matrix:
       - IMAGE: [amazonlinux:2, amazonlinux:2023, debian:10, debian:11, rockylinux:9.0, ubuntu:20.04, ubuntu:22.04]
-        INSTALL_CLASSIC_AGENT: ['', --opw-install-classic-agent install_script_agent7.sh]
+        DD_OPW_INSTALL_CLASSIC_AGENT: ['', install_script_agent7.sh]
   script:
-    - ./test/dockertest.sh --image "registry.ddbuild.io/images/mirror/${IMAGE}" --script install_script_op_worker1.sh --observability-pipelines-worker "true" ${INSTALL_CLASSIC_AGENT}
+    - ./test/dockertest.sh --image "registry.ddbuild.io/images/mirror/${IMAGE}" --script install_script_op_worker1.sh ${INSTALL_CLASSIC_AGENT}
```
KevinFairise2, line 183:
```suggestion
    - ./test/dockertest.sh --image "registry.ddbuild.io/images/mirror/${IMAGE}" --script install_script_op_worker1.sh
```

If successfuly figured out how all these variables are used I think we can remove this now, right?

KevinFairise2 (approved):
On small change other LGTM

.gitlab-ci.yml
chouquette: Indeed, I just amended the to do so!

chouquette:
I'm having 2nd thoughts about this change. I think the gitlab CI changes make sense, but we shouldn't touch the dockertest script as it can still be used by developers locally to test their changes in a container

PR #121, opened 2023-11-22
CI: migrate test runs to k8s runners [APL-2076]

chouquette:


test/dockertest.sh
chouetz, line 1: Do we need to keep this file?

chouetz (approved):

test/dockertest.sh
chouquette: I was planning on removing it as well but it was mentionned in the readme as the way for developer to test their changes so I kept, it but I can be convinced otherwise :)
It's of no use for the CI in any case though

PR #122, opened 2023-11-22
CI: move datadog-agent variable handling to a common before_script

chouquette:
This handling needs to be done for all jobs, not only the one for the datadog-agent test.


amenasria (approved):
LGTM!

PR #123, opened 2023-11-23
CI: provide a default value for MAJOR_VERSION

chouquette:
When the pipeline is triggered from another project, we don't include the MAJOR_VERSION from the test matrices, which cause the repo URL to be incorrect and the install to fail

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -10,11 +10,12 @@ variables:
   DATADOG_AGENT_BUILDIMAGES: v13321514-644fed4
   S3_CP_CMD: aws s3 cp --only-show-errors --region us-east-1 --sse AES256
   TEST_INFRA_DEFINITIONS_BUILDIMAGES: cb9e86993fd5
+  DEFAULT_MAJOR_VERSION: 7
 
 default:
   # Handle inputs when running as a downstream pipeline from the datadog-agent repo
   before_script:
-    - '[ -z "$MAJOR_VERSION" ] && export MAJOR_VERSION=7'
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -14,6 +14,7 @@ variables:
 default:
   # Handle inputs when running as a downstream pipeline from the datadog-agent repo
   before_script:
+    - '[ -z "$MAJOR_VERSION" ] && export MAJOR_VERSION=7'
```
chouetz, line 17: nitpick, should the default value be saved as a variable, for more flexibility?

chouetz (approved):

.gitlab-ci.yml
chouquette: that makes sense, I'll update

.gitlab-ci.yml
```diff
@@ -10,10 +10,12 @@ variables:
   DATADOG_AGENT_BUILDIMAGES: v13321514-644fed4
   S3_CP_CMD: aws s3 cp --only-show-errors --region us-east-1 --sse AES256
   TEST_INFRA_DEFINITIONS_BUILDIMAGES: cb9e86993fd5
+  DEFAUT_MAJOR_VERSION: 7
```
KevinFairise2, line 13: **🥜 nitpick:** ‏
```suggestion
  DEFAULT_MAJOR_VERSION: 7
```

chouquette: oh my :worried: 
Thanks for spotting, fixed

PR #139, opened 2024-01-23
[E2E] Install script e2e tests are fetching the latest available python version

Pythyu:
Add the `getLatestEmbeddedPythonPath` function to fetch the path to the latest python version embedded into an agent

### DIFF
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -21,6 +21,7 @@ import (
 	version "github.com/hashicorp/go-version"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+	yaml "gopkg.in/yaml.v2"
 )
 
 type osConfig struct {
--- a/test/e2e/common.go
+++ b/test/e2e/common.go
@@ -110,8 +110,10 @@ func getEC2Options(t *testing.T) []ec2params.Option {
 
 func (s *linuxInstallerTestSuite) getLatestEmbeddedPythonPath(baseName string) string {
 	path := fmt.Sprintf("/opt/%s/embedded/lib/python*", baseName)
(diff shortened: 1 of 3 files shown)


test/e2e/common.go
```diff
@@ -107,6 +107,34 @@ func getEC2Options(t *testing.T) []ec2params.Option {
 	return ec2Options
 }
 
+func (s *linuxInstallerTestSuite) getLatestEmbeddedPythonPath(baseName string) string {
+	vm := s.Env().VM
```
pducolin, line 111: **🥜 nitpick**
```suggestion
	s.T().Helper()
	vm := s.Env().VM
```

pducolin (approved):

test/e2e/common.go
```diff
@@ -145,7 +173,7 @@ func (s *linuxInstallerTestSuite) addExtraIntegration() {
 	t.Log("Install an extra integration, and create a custom file")
 	_, err := vm.ExecuteWithError("sudo -u dd-agent -- datadog-agent integration install -t datadog-bind9==0.1.0")
 	assert.NoError(t, err, "integration install failed")
-	_ = vm.Execute(fmt.Sprintf("sudo -u dd-agent -- touch /opt/%s/embedded/lib/python3.11/site-packages/testfile", s.baseName))
+	_ = vm.Execute(fmt.Sprintf("sudo -u dd-agent -- touch %s/site-packages/testfile", s.getLatestEmbeddedPythonPath(s.baseName)))
```
pducolin, line 176: **❓ question**
Aren't we missing a test to validate we have the embedded version we expect ? This would work on updates, but what if we fail installing the expected python version ?

pducolin: Will move the minor python version check on `datadog-agent` installer tests

PR #140, opened 2024-01-30
Invalidate install.datadoghq.com distribution on script release

bkabrda:
I tested the invalidation in another [testing branch](https://github.com/DataDog/agent-linux-install-script/compare/slavek.kabrda/test-cf-invalidate?expand=1) (because the only way to test it fully would be to do an actual release).

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,6 +6,7 @@ stages:
   - deploy
 
 variables:
+  AWS_MAX_ATTEMPTS: 5  # retry AWS operations 5 times if they fail on network errors
   DATADOG_AGENT_BUILDERS: v9930706-ef9d493
   DATADOG_AGENT_BUILDIMAGES: v13321514-644fed4
   S3_CP_CMD: aws s3 cp --only-show-errors --region us-east-1 --sse AES256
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -368,3 +369,15 @@ deploy:
       - SCRIPT: install_script_vector0.sh
   script:
     - $S3_CP_CMD ./${SCRIPT} s3://dd-agent/scripts/${SCRIPT} --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers full=id=3a6e02b08553fd157ae3fb918945dd1eaae5a1aa818940381ef07a430cf25732
+  after_script:
+    # invalidate the install.datadoghq.com CF distribution
+    - export $(printf "AWS_ACCESS_KEY_ID=%s AWS_SECRET_ACCESS_KEY=%s AWS_SESSION_TOKEN=%s"
```
chouquette, line 374: Can you actually do multiline commands like this?! I thought you either had to use `- >` or `- |` or at least enclose the string in a single/double quote :o
What a wonderful and clear syntax :sweat_smile:

.gitlab-ci.yml
```diff
@@ -368,3 +369,15 @@ deploy:
...
+        $(aws --region "us-east-1" sts assume-role
+          --duration-seconds 900
+          --role-arn "arn:aws:iam::464622532012:role/build-stable-cloudfront-invalidation"
+          --role-session-name "build-stable-cloudfront-invalidate-script"
+          --query "Credentials.[AccessKeyId,SecretAccessKey,SessionToken]"
+          --output text
+        )
+      )
+    - aws --region "us-east-1" cloudfront create-invalidation --distribution-id "E2VSER0FO39KRV" --paths "/scripts/*"
```
chouquette, line 383: This looks good, although out of curiosity, can't the same be achieved in a simpler way with `aws-vault`?

chouquette (approved):

.gitlab-ci.yml
bkabrda: Yes you can, the yaml parser will connect them into a single oneliner string. The `- >` and `- |` are for when you need multiple lines.

.gitlab-ci.yml
bkabrda: It can be achieved with aws-vault, but then you would have to create vault config file in a way very similar to how we create the environment variables, so basically there would be little difference.

PR #141, opened 2024-01-30
Enforce retries in install_script for curl calls

chouetz:
Add options in curl and APT to have more resilience in install_script.
`yum` as a [default retry to 10](https://github.com/Distrotech/yum/blob/f4e54aeed297158c563828aa3ebb93d0c8ce7e38/docs/yum.conf.5#L364-L366) and [no option exist yet](https://github.com/openSUSE/zypper/issues/420) in zypper

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -62,7 +62,7 @@ useful and we will do our very best to help you solve your problem.\n"
 }
 
 function report(){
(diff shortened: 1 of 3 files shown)


spencergilbert (approved):
Would you mind documenting the yum and zypper information in the code, otherwise LGTM

PR #142, opened 2024-02-09
Add install script for OPW v2

neuronull:
In OP / OPW 2.0 we will no longer support local pipeline configurations- remote configuration is the default. As such, the logic around remote configuration env vars and local pipelines needs to be removed from the install script.

https://datadoghq.atlassian.net/browse/OPA-1090

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -39,6 +39,7 @@ generate-scripts:
       - install_script_agent7.sh
       - install_script_docker_injection.sh
       - install_script_op_worker1.sh
+      - install_script_op_worker2.sh
       - install_script_vector0.sh
 
 unit_tests:
@@ -197,7 +198,7 @@ test-observability-pipelines-worker:
   stage: test
   dependencies: ["generate-scripts"]
   variables:
-    SCRIPT: ./install_script_op_worker1.sh
+    SCRIPT: ./install_script_op_worker2.sh
   parallel:
     matrix:
       - IMAGE:
@@ -366,6 +367,7 @@ deploy:
       - SCRIPT: install_script_agent7.sh
       - SCRIPT: install_script_docker_injection.sh
       - SCRIPT: install_script_op_worker1.sh
+      - SCRIPT: install_script_op_worker2.sh
       - SCRIPT: install_script_vector0.sh
   script:
     - $S3_CP_CMD ./${SCRIPT} s3://dd-agent/scripts/${SCRIPT} --grants read=uri=http://acs.amazonaws.com/groups/global/AllUsers full=id=3a6e02b08553fd157ae3fb918945dd1eaae5a1aa818940381ef07a430cf25732
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -198,7 +198,8 @@ test-observability-pipelines-worker:
   stage: test
   dependencies: ["generate-scripts"]
   variables:
-    SCRIPT: ./install_script_op_worker2.sh
+    # TODO once OPW 2.0 has a production release this needs to change to the v2 script
+    SCRIPT: ./install_script_op_worker1.sh
   parallel:
     matrix:
       - IMAGE:
--- a/install_script_op_worker1.sh
+++ b/install_script_op_worker1.sh
@@ -159,6 +159,11 @@ if [ -n "$DD_OP_PIPELINE_ID" ]; then
     op_pipeline_id=$DD_OP_PIPELINE_ID
 fi
 
+op_rc_enabled=
+if [ -n "$DD_OP_REMOTE_CONFIGURATION_ENABLED" ]; then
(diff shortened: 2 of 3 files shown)


spencergilbert:
👋 do we know how we'd like the UX to be when upgrading from OPW v1 -> v2? Should the file get moved to a non-referenced location/deleted/ignored/etc?

neuronull:
> 👋 do we know how we'd like the UX to be when upgrading from OPW v1 -> v2? Should the file get moved to a non-referenced location/deleted/ignored/etc?

Good question. There hasn't been discussion on this point that I'm aware of yet.

The worker will be ignoring the pipeline configuration file. If it is there it doesn't cause any harm. The data_dir is still being used for other things. IMO the least invasive approach is to just leave that file alone. Users may have information in there they want to preserve.

spencergilbert:
I suspect we can do some dancing around the version in script like we have [here](https://github.com/DataDog/agent-linux-install-script/blob/main/install_script.sh.template#L655-L674).

If the major version is `1` we can keep doing the script you've removed currently, otherwise ignore it. Do you mind updating the PR to do that?

neuronull:
> If the major version is 1 we can keep doing the script you've removed currently, otherwise ignore it. Do you mind updating the PR to do that?

Makes sense.

Looking at how the major version is determined, it seems we rely on an env var and default to 1 if it is not set ... Do you know how/where that env var is expected to be set? As it stands it looks like we are relying on the default value, which wouldn't work if we're planning to use the same script for both OP 1 and OP 2.

spencergilbert:
IIRC, that is optionally passed in when calling the install script. I was thinking we would do that now as prep for the 2.0 release, and it would allow y'all to test the installer by passing the upcoming version.

Once 2.0 is released we could update the default version and have users opt-in to the old behavior if we're allowing them to keep using 1.x for some amount of time, or if 1.x is going to be unsupported by the backend we can do a harder removal.

A bigger change would be to adopt the template strategy the agent install script has and have it create the `install_script_op_worker1.sh` and `install_script_op_worker2.sh` based on the supported features.

neuronull:
Roger 👍 

We would support 1.x for some amount of time. I suspect in the references to this script in the 1.x versions of the app, we could preload the env var as part of the command that is referenced in the app. I suspect that will be fine but I can double check.

spencergilbert:
I'm doublechecking with the team tomorrow if we'd prefer that or the versioned script 👍

neuronull:
```
diff install_script_op_worker1.sh install_script_op_worker2.sh
```

```diff
160,164d159
< fi
<
< op_rc_enabled=
< if [ -n "$DD_OP_REMOTE_CONFIGURATION_ENABLED" ]; then
<     op_rc_enabled=$DD_OP_REMOTE_CONFIGURATION_ENABLED
224d218
< pipeline_file="$etcdir/pipeline.yaml"
227c221
< worker_major_version=1
---
> worker_major_version=2
484,488d477
<
<   if [ "$op_rc_enabled" ]; then
<     printf "\033[34m  * Assigning DD_OP_REMOTE_CONFIGURATION_ENABLED.\n\033[0m\n"
<     $sudo_cmd sh -c "echo DD_OP_REMOTE_CONFIGURATION_ENABLED=$op_rc_enabled >> $env_file"
<   fi
503,510d491
< if [ ! -e "$pipeline_file" ]; then
<   if ! $sudo_cmd grep -q -E '^DD_OP_REMOTE_CONFIGURATION_ENABLED=true|"true"' "$env_file" && \
<     ! $sudo_cmd grep -A1 -E '^remote_configuration:' "$bootstrap_file" | $sudo_cmd grep -q -E '^\s{2}enabled: true|"true"'; then
<     printf "\033[33mThe $nice_flavor won't start automatically at the end of the script because the pipeline configuration is missing.\n  If you are following our onboarding guide, please return to the Datadog Observability Pipelines UI to download your pipeline's configuration.\n\033[0m\n"
<     no_start=true
<   fi
< fi
<
```

bkabrda (changes requested):
The script itself LGTM, but could you please add a new entry for this script in `.gitlab-ci.yml` (`deploy` job) so we would have a job to actually release the script?

neuronull:
> The script itself LGTM, but could you please add a new entry for this script in `.gitlab-ci.yml` (`deploy` job) so we would have a job to actually release the script?

Ah, yes. I believe this is covered now in: [acc34b7](https://github.com/DataDog/agent-linux-install-script/pull/142/commits/acc34b70d1764c831dedb453cc87c346e3e34d34)

bkabrda (approved):

bkabrda:
@neuronull I approved, but immediately realized that there's a problem with the `test-observability-pipelines-worker` job - this won't work until we have a prod release of OPW 2. Can you please revert that change to use OPW 1 still? We can flip it to OPW 2 once you guys make the first prod release. (The CI doesn't run on your changes because you're making changes from fork and that doesn't trigger our internal Gitlab CI)

spencergilbert (approved):

PR #143, opened 2024-02-22
Add option to install updater

raphaelgavache:
Updater install is at the time of the PR only available on apt testing pipelines


chouquette (approved):
LGTM
You might want to add a few more checks to ensure that the updater installation is mutually exclusive with other flavors/modes

PR #147, opened 2024-02-27
fix: Update makefile and script metadata

spencergilbert:
- **fix: Add op_worker2 script to Makefile release targets**
- **fix: Update op_worker2 metadata**

These two places were missed when the v2 script was added recently.


bkabrda (approved):
Thanks, I forgot about these when reviewing the OPW 2 script :/

PR #148, opened 2024-02-29
chore: Prepare 1.27.1 release

spencergilbert:


chouetz (approved):

CHANGELOG.rst
```diff
@@ -2,9 +2,12 @@
 Release Notes
 =============
 
-Unreleased
+1.27.1
 ================
 
+- Fix OPW v2 script metadata (#147)
+- Update buildimages tag (#146)
+
```
KSerrania, line 10: Thought for later: should we separate functional changes to the install script, from tooling / internal changes in the changelog?

KSerrania (approved):

CHANGELOG.rst
spencergilbert: Yeah, I wasn't entirely sure it needed to be included - but since we had some "plumbing" notes in previous releases I opted to include everything.

I don't know how frequently referenced the changelog is here, but it does seem like an easy improvement to separate out user-facing notes.

PR #149, opened 2024-03-05
Don't allow installing Agent > 6/7.51 on CentOS 6

bkabrda:
This will:
* Only allow installing pinned versions <= 6/7.51 on CentOS/RHEL 6
* Install 6/7.51 when performing an "unpinned" installation.


Pythyu (approved):
LGTM 🚀

PR #152, opened 2024-03-06
fix: Fix permissions error when installing on CentOS

spencergilbert:
This PR has been manually tested, but does not include automated tests that continue to verify the expected behavior.


Pythyu (approved):
🚀

PR #157, opened 2024-03-12
Activate security products

robertpi:


test.sh
bkabrda, line 1: Was this file added by mistake? I don't see it being used anywhere.

install_script.sh.template
```diff
@@ -1419,6 +1436,8 @@ elif [ ! "$no_agent" ]; then
   manage_security_and_system_probe_config "$sudo_cmd" "$security_agent_config_file" "$system_probe_config_file" "$DD_RUNTIME_SECURITY_CONFIG_ENABLED" "$DD_COMPLIANCE_CONFIG_ENABLED"
 fi
 
+manage_client_libraries_security_config "$sudo_cmd" "$environment_file" "$DD_APPSEC_ENABLED" "$DD_IAST_ENABLED" "$DD_APPSEC_SCA_ENABLED"
```
bkabrda, line 1439: Nit: two variables names start with `DD_APPSEC_` and one only with `DD_`. Is this intentional? Would it make more sense to make them all start with `DD_APPSEC_`?

test.sh
robertpi: Gosh, that's a silly mistake. Remove it in bb31bb0

install_script.sh.template
robertpi: While I agree with your comments, these settings have be implemented and document in the client libraries for quite sometime, so are not possible to change without requesting all customers using them to migrate.

unit_tests/test_install_script.sh
```diff
@@ -249,5 +250,27 @@ testFullConfig(){
   assertEquals 1 $?
 }
 
+### Manage client libraries security config
+testManageClientLibrariesSecurityConfig() {
+  rm $environment_file
```
spencergilbert, line 255: [Small note that this sort of errors, or at least just emits an unneeded log](https://gitlab.ddbuild.io/DataDog/agent-linux-install-script/-/jobs/460233248#L512) - don't expect it to be a blocker/problem, just unexpected for me to see it when I was reviewing the test logs.

spencergilbert (approved):

PR #158, opened 2024-03-13
fix: Update Makefile targets to zero out patch version on minor update, and correctly generate changelog when version is specified

spencergilbert:


Pythyu (approved):
LGTM

PR #163, opened 2024-03-14
Make ubuntu updater test work

raphaelgavache:
The ubuntu tests failed as systemd units entered a failed state due to missing datadog.yaml. Having a datadog.yaml before installing the updater is necessary to ensure that the updater knows what to install.
Customers can set rules such as: install agent version 7.51 on env prod

Fixes
1. Create an early datadog.yaml in the case of the updater
2. Ignore services restart on updater installation


raphaelgavache:
does not work, creating the first config file part isn't working

PR #169, opened 2024-03-15
Small but important correction

robertpi:


install_script.sh.template
```diff
@@ -651,7 +651,7 @@ fi
 config_file_fips=$etcdirfips/datadog-fips-proxy.cfg
 system_probe_config_file=$etcdir/system-probe.yaml
 security_agent_config_file=$etcdir/security-agent.yaml
-environment_file=$etcdir/environment
+environment_file=/etc/environment
```
spencergilbert, line 654: **💬 suggestion:** ‏ Since this variable is only used in one place, why not just pass the `/etc/environment` value directly?

robertpi: okay, done in af52838

spencergilbert (approved):

PR #170, opened 2024-03-15
[BARX-289] fix updater ubuntu test + bootstrap config

raphaelgavache:
The ubuntu tests failed as systemd units entered a failed state due to missing datadog.yaml. Having a datadog.yaml before installing the updater is necessary to ensure that the updater knows what to install.
Customers can set rules such as: install agent version 7.51 on env prod

Fixes

Create an early datadog.yaml in the case of the updater
Ignore services restart on updater installation

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -418,9 +418,29 @@ if [ -n "$DD_AGENT_FLAVOR" ]; then
     agent_flavor=$DD_AGENT_FLAVOR #Eg: datadog-iot-agent
 fi
 
+# Root user detection
+if [ "$UID" == "0" ]; then
+    sudo_cmd=''
+else
+    sudo_cmd='sudo'
+fi
+
 install_updater=
 if [ -n "$DD_INSTALL_UPDATER" ]; then
     install_updater=true
+    config_file="/etc/datadog-agent/datadog.yaml"
+    if [ ! -e $config_file ]; then
+        $sudo_cmd mkdir -p /etc/datadog-agent
+        $sudo_cmd touch $config_file
+        $sudo_cmd chmod 666 $config_file # very bad practice, but unlocks internal trials
+        echo "api_key: $apikey" > $config_file
+        echo "site: $site" >> $config_file
+	echo "hostname: $hostname" >> $config_file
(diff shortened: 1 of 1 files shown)


install_script.sh.template
```diff
@@ -418,9 +418,29 @@ if [ -n "$DD_AGENT_FLAVOR" ]; then
     agent_flavor=$DD_AGENT_FLAVOR #Eg: datadog-iot-agent
 fi
 
+# Root user detection
```
raphaelgavache, line 421: going to cleanup it and move to 1075

spencergilbert (commented):
I think my biggest concern about this strategy, is how it interacts with the following configuration steps, and conditionals around _not_ touching the config file if it already exists.

install_script.sh.template
```diff
@@ -1053,6 +1053,20 @@ If the cause is unclear, please contact Datadog support.
...
+        $sudo_cmd mkdir -p /etc/datadog-agent
+        $sudo_cmd touch $config_file
+        $sudo_cmd chmod 644 $config_file
+        $sudo_cmd sh -c "echo 'api_key: $apikey' > $config_file"
+        $sudo_cmd sh -c "echo 'site: $site' >> $config_file"
+        $sudo_cmd sh -c "echo 'hostname: $hostname' >> $config_file"
+        formatted_host_tags="['""$( echo "$host_tags" | sed "s/,/', '/g" )""']"  # format `env:prod,foo:bar` to yaml-compliant `['env:prod','foo:bar']`
+        $sudo_cmd sh -c "echo 'tags: $formatted_host_tags' >> $config_file"
+        $sudo_cmd sh -c "echo 'env: $env' >> $config_file"
+        $sudo_cmd sh -c "echo 'dd_url: $url' >> $config_file"
```
spencergilbert, line 1068: **❓ question:** ‏ Are all of these required for the updater? We discussed some in our standup and it was pointed out most of the configuration would be done from our UI - using remote config, right?

raphaelgavache: Yes all of them would be useful.
- apiKey is used for auth
- env + host tags + hostnames are used for configuration routing, we will match on these to know fetch the installation configration
- dd_url/site are used to target the correct datadog

spencergilbert (approved):

PR #174, opened 2024-03-29
Install Old 2022 RPM GPG key only when needed for agent <= 7.35

Pythyu:
Pop the last GPG key (expired in 2022) from the `RPM_GPG_KEYS` to install if we're not using an agent <= 7.35

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -300,6 +300,17 @@ function remove_rpm_gpg_keys() {
     done
 }
 
+function get_relevant_gpg_keys() {
+    local minorVersion="$1"
+    shift
+    local gpgKeys=("$@")
+    # The last GPG key is only used on agent 7.35.X and lower
+    if [ "minorVersion" -gt 35 ]; then
+        gpgKeys=("${gpgKeys[@]/$gpgKeys[3]}")
+    fi
+    echo "${gpgKeys[@]}"
+}
+
 # Emulate hashmap with simple switch case
 function getMapData() {
 
@@ -876,7 +887,7 @@ if [ "$OS" == "RedHat" ]; then
 
     gpgkeys=''
     separator='\n       '
-    for key_path in "${RPM_GPG_KEYS[@]}"; do
+    for key_path in $(get_relevant_gpg_keys "$agent_minor_version" "${RPM_GPG_KEYS[@]}"); do
       gpgkeys="${gpgkeys:+"${gpgkeys}${separator}"}https://${keys_url}/${key_path}"
     done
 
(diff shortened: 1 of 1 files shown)


install_script.sh.template
```diff
@@ -300,6 +300,17 @@ function remove_rpm_gpg_keys() {
     done
 }
 
+function get_relevant_gpg_keys() {
+    local minorVersion="$1"
+    shift
```
spencergilbert, line 305: **❓ question:** ‏What does `shift` do here?

spencergilbert (commented):
Is this something we can add at test for?

install_script.sh.template
```diff
@@ -300,6 +300,17 @@ function remove_rpm_gpg_keys() {
...
+    local gpgKeys=("$@")
+    # The last GPG key is only used on agent 7.35.X and lower
+    if [ -z "$minorVersion" ] || [ "$minorVersion" -gt 35 ]; then
+        unset 'gpgKeys[${#gpgKeys[@]}-1]'
```
chouquette, line 309: I fear this might bite us in the future, if we ever remove the key from the initial array.

Do you think we could process the array once, ideally immediately after defining it?

Pythyu: 🤦 I think my description was scuffed, I need to check but I think it's only popping from the local array `gpgKeys`

install_script.sh.template
Pythyu: `shift` move forward the argument list by one meaning that in arguments of the function I have `get_relevant_gpg_keys  version gpgkey1 gpgkey2 gpgkey3 gpgkey4` after shift it's going to be as `get_relevant_gpg_keys gpgkey1 gpgkey2 gpgkey3 gpgkey4` and `local gpgKeys=("$@")` will take every remaining arguments inside of a new array `gpgKeys`

install_script.sh.template
chouquette: Sure, but it would still stop installing/using the oldest key. It currently works fine because the oldest key is the one we want to remove for this version, but this won't hold if the list of keys change

install_script.sh.template
chouquette: I need to refresh pages before answering to already answered question :sweat_smile:

install_script.sh.template
```diff
@@ -300,6 +300,17 @@ function remove_rpm_gpg_keys() {
     done
 }
 
+function get_relevant_gpg_keys() {
+    local minorVersion="$1"
+    shift
+    local gpgKeys=("$@")
+    # The last GPG key is only used on agent 7.35.X and lower
+    if [ -z "$minorVersion" ] || [ "$minorVersion" -gt 35 ]; then
```
chouquette, line 308: Do we really want to remove the key when the minor version isn't set?

chouquette: Ah ok got it! Then it's fine indeed :)

install_script.sh.template
Pythyu: Ah yes but when we're doing the gpg key rotation in june we'll just have to add or something like that
```
if [ -z "$minorVersion" ] || [ "$minorVersion" -gt 52 ]; then
        unset 'gpgKeys[${#gpgKeys[@]}-2]'
```
inside of the same function no ?

install_script.sh.template
spencergilbert: Thanks!

spencergilbert (approved):

PR #175, opened 2024-04-14
(fleet) install the installer on Debian when APM is enabled

arbll:
This PR:
- Installs the datadog-installer on Debian when APM is enabled
- Adds telemetry specific to the installer
- Populates `/etc/datadog-agent/datadog-installer.ini` with wether APM is enabled and with traceID/spanID for telemetry


install_script.sh.template
```diff
@@ -378,6 +378,212 @@ function ensure_config_file_exists() {
...
+    echo -e "\033[34m\n* Reporting installer telemetry\n\033[0m"
+
+    if [ "$DD_INSTRUMENTATION_TELEMETRY_ENABLED" == "false" ] || \
+      [ "$site" == "ddog-gov.com" ] || \
+      [ -z "${apikey}" ] || \
+      [ -z "$telemetry_url" ]; then
+      return
+    fi
+
+    time_now_seconds=$(date +%s)
```
raphaelgavache, line 408: if any of those command fail, wont the above `set -e` trigger an end of script?
in which case a || defaultValue would protect or maybe wrapping the install command with set +e, set -e

install_script.sh.template
```diff
@@ -378,6 +378,212 @@ function ensure_config_file_exists() {
...
+            ]]
+        }
+    }
+END
+)
+
+    printf -v log_lines_stderr "%b" "$install_stderr"
+    printf -v log_lines_stdout "%b" "$install_stdout"
+    json_logs="["
+    while IFS= read -r line; do
```
raphaelgavache, line 466: to avoid surprises you can set back to previous IFS value
add a OLD_IFS=$IFS and set it back after

arbll: I don't think it will override it, it's the usual `ENV_VAR command`, not an assignment

install_script.sh.template
arbll: Isolated the telemetry in a subshell to nullify the risk

install_script.sh.template
```diff
@@ -378,6 +378,214 @@ function ensure_config_file_exists() {
...
+EOT"
+}
+
+function _install_installer() {
+    local sudo_cmd="$1"
+    local os="$2"
+    local distribution="$3"
+    local apm_instrumentation_enabled="$4"
+
+    if [ "$os" == "Debian" ] && [ "$distribution" == "Debian" ] && [ -n "$apm_instrumentation_enabled" ]; then
```
chouquette, line 555: `OS` is normalized to our handled distribution, you can probably remove the comparison on `$distribution`

install_script.sh.template
```diff
@@ -378,6 +378,214 @@ function ensure_config_file_exists() {
...
+        echo -e "\033[34m\n* Installing the Datadog installer\n\033[0m"
+        configure_installer "$sudo_cmd" "$trace_id" "$apm_instrumentation_enabled"
+        $sudo_cmd apt-get install -o Acquire::Retries="5" -y --force-yes "datadog-installer"
```
chouquette, line 558: is `--force-yes` really needed? It's deprecated and causes a bunch of warnings to be emitted during each test

install_script.sh.template
```diff
@@ -378,6 +378,214 @@ function ensure_config_file_exists() {
...
+
+    trace_id=$(od -An -N8 -tu8 < /dev/urandom | tr -d ' ')
+    start_time=$(date +%s%N)
+
+    (_install_installer "$sudo_cmd" "$os" "$distribution" "$apm_instrumentation_enabled" > /tmp/datadog-installer-stdout.log 2> /tmp/datadog-installer-stderr.log) || exit_status=$?
+    exit_status=${exit_status:-0}
+
+    # escape newlines and double quotes before sending the telemetry
+    install_stdout=$(cat /tmp/datadog-installer-stdout.log | sed ':a;N;$!ba;s/\n/\\n/g' | sed 's/"/\\"/g')
+    install_stderr=$(cat /tmp/datadog-installer-stderr.log | sed ':a;N;$!ba;s/\n/\\n/g' | sed 's/"/\\"/g')
```
chouquette, line 583: This seems frail, maybe using `jq` to compose the telemetry payload would be more robust

install_script.sh.template
arbll: I copy / pasted what we had for the other DEB packages, I don't actually know if we need this, it creates the same logs for the regular agent install 😅

chouquette: It probably should be removed from there as well :)

install_script.sh.template
arbll: I initially discarded that option because we'd need to add it as a dependency to install and I'm not sure it's going to have a package available in all the distros we support. Do you think it's fine to add it as a dep ?

install_script.sh.template
arbll: I initially wanted debian specifically and not ubuntu to limit as much as possible the scope but we might revisit. (AFAICT it's possible to have $os/$distribution be either debian/debian or debian/ubuntu if I read https://github.com/DataDog/agent-linux-install-script/blob/a65b0c07a6a8c11a105c13dfced1f326d8a10195/install_script.sh.template#L794-L795 properly)

install_script.sh.template
arbll: Ok I'll remove it here and we can use this as a test that it's safe to remove :trollface:

chouquette: Hm actually I might have been too eager to remove it, it's not supported by ubuntu 14.04, which we still support :(

arbll: classic, adding it back lol

install_script.sh.template
chouquette: Ah ok gotcha, then all good

install_script.sh.template
chouquette: Sorry :sweat_smile:

chouquette (approved):

raphaelgavache (approved):

PR #177, opened 2024-04-18
Add missing env vars to OP 2.0 script

neuronull:
During OP 2.0 development in Q1 2024 we added a number of env vars that need to be supported in the install script.

This is a first pass hard coded solution.

### DIFF
--- a/install_script_op_worker2.sh
+++ b/install_script_op_worker2.sh
@@ -151,12 +151,12 @@ fi
 
 apikey=
 if [ -n "$DD_API_KEY" ]; then
-    apikey=$DD_API_KEY
+    apikey="$DD_API_KEY"
 fi
 
 op_pipeline_id=
 if [ -n "$DD_OP_PIPELINE_ID" ]; then
-    op_pipeline_id=$DD_OP_PIPELINE_ID
+    op_pipeline_id="$DD_OP_PIPELINE_ID"
 fi
 
 no_start=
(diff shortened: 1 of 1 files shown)


install_script_op_worker2.sh
```diff
@@ -475,6 +533,64 @@ else
     printf "\033[34m  * Assigning DD_SITE.\n\033[0m\n"
     $sudo_cmd sh -c "echo DD_SITE=$site >> $env_file"
   fi
+
+  # sources
+  if [ "$source_splunk_hec_address" ]; then
+    printf "\033[34m  * Assigning DD_OP_SOURCE_SPLUNK_HEC_ADDRESS.\n\033[0m\n"
+    $sudo_cmd sh -c "echo DD_OP_SOURCE_SPLUNK_HEC_ADDRESS=$source_splunk_hec_address>> $env_file"
+  fi
```
spencergilbert, line 541: **❓ question:** ‏I think this is just my brain not quite working fully - but with `source_splunk_hec_address=`, meaning the environment variable isn't set, this conditional fails and we don't write it to the file?

install_script_op_worker2.sh
spencergilbert, line 1: **❗  note:** ‏This was discussed over Zoom, but for posterity - given a consistent prefix on the environment variables we could look to dynamically write the passed variables to the env file. I noted a possible concern about unintentionally writing a value that _shouldn't_ be saved, this feels acceptable given we already write sensitive values to the file.

Dynamically writing any/all prefixed variables would future proof the script as we include additional components to OPW. If we hard code the values we'll be duplicating some code, as well as requiring future PRs and releases when more component support is added.

install_script_op_worker2.sh
```diff
@@ -164,6 +164,64 @@ if [ -n "$DD_INSTALL_ONLY" ]; then
...
+if [ -n "$DD_OP_SOURCE_DATADOG_AGENT_ADDRESS" ]; then
+    source_datadog_agent_address=$DD_OP_SOURCE_DATADOG_AGENT_ADDRESS
+fi
+
+source_sumo_logic_address=
+if [ -n "$DD_OP_SOURCE_SUMO_LOGIC_ADDRESS" ]; then
+    source_sumo_logic_address=$DD_OP_SOURCE_SUMO_LOGIC_ADDRESS
+fi
+
+# source key passes
```
spencergilbert, line 188: **❓ question:** ‏Just confirming these names are correct because KEY_PASS sounds funny.

spencergilbert (approved):
It looks like anything that could be wrong would be a typo, but I didn't catch anything. I think it could be worth adding an e2e test ensuring everything is written in and read by OPW correctly, but that can be a follow-up task.

install_script_op_worker2.sh
neuronull: yep its a bit weird but those ones are correct

install_script_op_worker2.sh
neuronull: 👍 

as I was writing this up I also had the thought that we could make this slightly less duplicate code by having some helper functions do some of this for us, a static list of the env vars for the components etc. 

so its not really dynamic but less copy pasta

spencergilbert: Same thing came to my mind as I was reviewing, great minds and all that. Possibly worth combining the two and being able to report to the user "you added this DD_OP_* var, but that doesn't match a known component" - could help catch typos and misconfigurations ahead of time.

install_script_op_worker2.sh
neuronull: Correct, we dont write to the $env_file for the respective env var, if the the relevant DD_OP_* env var is not set

From the customer's use case, they will be presented with a snippet to copy-paste and run. And the relevant env vars they need based on their pipeline are prepended to the install script invocation.

neuronull:
>  I think it could be worth adding an e2e test ensuring everything is written in and read by OPW correctly, but that can be a follow-up task.

Agreed. I can open a ticket to track that and also one to track circling back to a more robust solution here

install_script_op_worker2.sh
spencergilbert: 👍 for a second I couldn't remember if `[ "$foo" ]` was true if the variable was set at all, and if "empty" counted as set. Bash is fun sometimes 😉

install_script_op_worker2.sh
```diff
@@ -164,6 +164,64 @@ if [ -n "$DD_INSTALL_ONLY" ]; then
     no_start=true
 fi
 
+# sources
+source_splunk_hec_address=
+if [ -n "$DD_OP_SOURCE_SPLUNK_HEC_ADDRESS" ]; then
+    source_splunk_hec_address=$DD_OP_SOURCE_SPLUNK_HEC_ADDRESS
+fi
```
bruceg, line 171: Maybe I'm missing something, as I haven't been watching these scripts. What does this construction achieve that a simple assignment doesn't?

Also note, since the RHS is unquoted, this script will misbehave if it contains spaces or other special characters.

install_script_op_worker2.sh
bruceg: It represents the password used to encrypt the TLS secret key data.

install_script_op_worker2.sh
bruceg: This should probably use `-n` to test if the variable is not empty, avoiding another trap of variables values starting with a hyphen causing confusion.

install_script_op_worker2.sh
spencergilbert: As far as I'm aware, a lot of "cargo culting" from the existing implementations. Even if it's unneeded I'd opt to keep it in-line with the other usages, and we can revisit with a larger fix/refactor.

neuronull: I'm not the OG author but after reviewing the script, I suspect this was intentionally done for things like the DD_API_KEY and pipeline_id , so that it would be easier to pass around the shorter local script var (apikey, op_pipeline_id) , than check the actual env var each place its used.

Because the component env vars that are being added here are not used in as many places, I believe we could remove this extraction to local vars and just use direct down circa line 538.

HOWEVER- there is talk in the slack thread about trying to get this done today... so if this works, in the sake of time I think we should proceed since we are planning to open a ticket to follow up with a more robust solution anyway.

WDYT?

I can add quotes to the RHS

bruceg (approved):
I confirm this covers all the env vars from the schema.

bruceg (approved):

PR #178, opened 2024-04-18
Revert "Install Old 2022 RPM GPG key only when needed for agent <= 7.35"

spencergilbert:
Reverts DataDog/agent-linux-install-script#174


spencergilbert:
Merging without approval due to IR-26823

PR #196, opened 2024-05-02
[E2E] Disable extraintegration check on pinned version agent

Pythyu:
integrations metadata expired since First May 2024 for 7.42.0 and needs to be disabled

Left as comment for now, might replace the test later with another test with `--unsafe-disable-verification` since we're pinning to the very first version of the package

### DIFF
--- a/test/e2e/install_test.go
+++ b/test/e2e/install_test.go
@@ -49,13 +49,14 @@ func (s *installTestSuite) TestInstall() {
 }
 
 func (s *installTestSuite) TestInstallMinorVersionPin() {
+	vm := s.Env().VM
+
 	// Installation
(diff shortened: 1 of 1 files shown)


test/e2e/install_test.go
```diff
@@ -54,7 +54,8 @@ func (s *installTestSuite) TestInstallMinorVersionPin() {
 
 	s.assertPinnedInstallScript("7.42.0")
 
-	s.addExtraIntegration()
+	// Disabled : integrations expired since First May 2024 for 7.42.0
+	//s.addExtraIntegration()
```
alopezz, line 58: I think it's fine if we forget about checking this specific functionality since I don't think it's related to the system under test.

```suggestion
```

alopezz (approved):

PR #208, opened 2024-05-15
(fleet) filter out packages installed by the installer

arbll:
This PR:
- Moves the installation of the installer at the top after input resolution
- Adds a way to check if a package is installed by the installer
- Filter out packages installed by the installer when installing package manager packages
- Avoid running the setup script of the injector if its installed with the installer

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -555,6 +555,39 @@ function install_installer() {
     return $exit_status
 }
 
+is_installed_by_installer() {
+    if command -v datadog-installer >/dev/null 2>&1; then
+        # If 'installer' is available, use it to check if the package is installed
+        installer is-installed "$1"
+        local status=$?
+        if [ $status -eq 0 ]; then
+            return 0  # Package is installed
+        elif [ $status -eq 10 ]; then
+            return 1  # Package is not installed
+        else
+            return $status  # Some other error occurred
+        fi
+    else
+        return 1
+    fi
+}
+
+filter_packages_installed_by_installer() {
+    local -n arr=$1  # Using nameref to refer to the array passed by name
+    local not_installed_packages=()
+
+    for package in "${arr[@]}"; do
+        if ! is_installed_by_installer "$package"; then
+            not_installed_packages+=("$package")
+        else
+            echo -e "\033[34m\n* Skipping installation of $package as it is already installed by the installer\n\033[0m"
+        fi
+    done
+
+    # Copy the filtered array back to the original array
+    arr=("${not_installed_packages[@]}")
+}
+
 echo -e "\033[34m\n* Datadog INSTALL_SCRIPT_REPORT_VERSION_PLACEHOLDER install script v${install_script_version}\n\033[0m"
 
 ##
@@ -1029,6 +1062,9 @@ else
     sudo_cmd='sudo'
 fi
 
+# Install the installer
+install_installer "$sudo_cmd" "$OS" "$DISTRIBUTION" "$DD_INSTALLER" "$DD_APM_INSTRUMENTATION_ENABLED" || true
+
 ##
 # INSTALL THE NECESSARY PACKAGE SOURCES
 ##
@@ -1143,6 +1179,7 @@ if [ "$OS" == "RedHat" ]; then
     # packages can be empty if no_agent is set
     if [ ${#packages[@]} -ne 0 ]; then
       echo -e "  \033[33mInstalling package(s): ${packages[*]}\n\033[0m"
+      filter_packages_installed_by_installer packages
 
       # yum has a default retry of 10 https://github.com/Distrotech/yum/blob/f4e54aeed297158c563828aa3ebb93d0c8ce7e38/docs/yum.conf.5#L364-L366
       $sudo_cmd yum -y --disablerepo='*' --enablerepo='datadog' install $dnf_flag "${packages[@]}" 2> >(tee /tmp/ddog_install_error_msg >&2) || $sudo_cmd yum -y install $dnf_flag "${packages[@]}" 2> >(tee /tmp/ddog_install_error_msg >&2)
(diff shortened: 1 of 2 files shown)


install_script.sh.template
```diff
@@ -1140,8 +1174,12 @@ if [ "$OS" == "RedHat" ]; then
       done
     fi
 
+    # Install the installer
+    install_installer "$sudo_cmd" "$OS" "$DISTRIBUTION" "$DD_INSTALLER" "$DD_APM_INSTRUMENTATION_ENABLED" || true
```
raphaelgavache, line 1178: nit, having both in the same function, embedding telemetry of the packages filtering would be nice

ex on the install span
- packages_pre_filter
- packages_filtered

install_script.sh.template
```diff
@@ -555,6 +555,40 @@ function install_installer() {
...
+    for package in $(eval echo "\${${pkg_array_name}[@]}"); do
+        if ! is_installed_by_installer "$sudo_cmd" "$package"; then
+            not_installed_packages+=("$package")
+        else
+            echo -e "\033[34m\n* Skipping installation of $package as it is already installed by the installer\n\033[0m"
+        fi
+    done
+
+    # Copy the filtered array back to the original array
+    eval "$pkg_array_name=(\"\${not_installed_packages[@]}\")"
```
raphaelgavache, line 589: I tried without eval using [nameref `-n` but it's only for bash 4.3+](https://www.gnu.org/software/bash/manual/bash.html#Locale-Translation).
Which makes me realise that we need to be sure that the array syntaxes are compatible across bash versions

raphaelgavache (approved):
lgtm

install_script.sh.template
arbll: Yeah I tried nameref at first but it failed on Centos7 and other old platforms

install_script.sh.template
arbll: yeah that's a good one, added it

BaptisteFoy (approved):

install_script.sh.template
```diff
@@ -519,13 +523,48 @@ function _install_installer() {
...
+    local sudo_cmd="$1"
+    if command -v datadog-installer >/dev/null 2>&1; then
+        $sudo_cmd datadog-installer is-installed "$2"
+        local status=$?
+        if [ $status -eq 0 ]; then
+            return 0
+        elif [ $status -eq 10 ]; then # 10 means the package is not installed
+            return 1
+        else
+            return $status
```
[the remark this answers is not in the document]
arbll: I think we will want to fail on this condition in the future since we don't actually know if the package is installed or not

install_script.sh.template
```diff
@@ -369,6 +369,8 @@ function report_installer_telemetry() {
     local exit_code="$5"
     local install_stdout="$6"
     local install_stderr="$7"
+    local packages_to_install="$8"
+    local pacakges_to_install_after_installer="$9"
```
chouquette, line 373: nit: typo

install_script.sh.template
```diff
@@ -549,9 +588,16 @@ function install_installer() {
     (_install_installer "$sudo_cmd" "$os" "$distribution" > /tmp/datadog-installer-stdout.log 2> /tmp/datadog-installer-stderr.log) || exit_status=$?
     exit_status=${exit_status:-0}
 
+    local packages_to_install
+    local pacakges_to_install_after_installer
```
chouquette, line 592: same typo, but this one might actually be problematic since you refer to the correct variable name afterward

install_script.sh.template
```diff
@@ -549,9 +588,16 @@ function install_installer() {
...
+
+    packages_to_install=("${!pkg_array_name}")
+    filter_packages_installed_by_installer "$sudo_cmd" "$pkg_array_name"
+    packages_to_install_after_installer=("${!pkg_array_name}")
```
chouquette, line 596: Is this only computed for telemetry or is it somehow propagated outside of this function?

install_script.sh.template
```diff
@@ -1140,6 +1186,10 @@ if [ "$OS" == "RedHat" ]; then
       done
     fi
 
+    # Install the installer
+    # Note: this function will remove installed packages from the "packages" array
+    install_installer "$sudo_cmd" "$OS" "$DISTRIBUTION" "$DD_INSTALLER" "$DD_APM_INSTRUMENTATION_ENABLED" "packages" || true
```
chouquette, line 1191: not specific to this callsite, but why do you use the variable name instead of the variable itself? (Please forgive my lack of shell scripting knowledge if this is a common pattern :sweat_smile: )

arbll: Bash does not support passing reference or arrays as arguments. The usual way to deal with this is by passing the name of the variable containing the array and resolving the array in the function from its name. Agreed it's black magic for anyone not familiar with bash but it's the only way to avoid doing the filtering inline which I tried to avoid since we have 3 callsites

chouquette: TIL! Thanks for the explanation

install_script.sh.template
arbll: only used for telemetry

install_script.sh.template
```diff
@@ -519,13 +523,48 @@ function _install_installer() {
...
 function install_installer() {
     local sudo_cmd="$1"
     local os="$2"
     local distribution="$3"
     local datadog_installer="$4"
     local apm_instrumentation_enabled="$5"
+    local pkg_array_name="$6"
 
     local trace_id
     local start_time
```
chouquette, line 570: These seem unused

chouquette (approved):
LGTM!

install_script.sh.template
arbll: It is used line 567

PR #212, opened 2024-05-22
Add a check to forbid APM injection installation on redhat 6

paullegranddc:
Installation was already broken because the installer is linked to a newer version of libc, but we want to exit early rather than installing broken packages.

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -1057,6 +1057,13 @@ if [ "$OS" == "RedHat" ]; then
                 exit;
             fi
         else
+            if [ ${#apm_libraries[@]} -gt 0 ]; then
+              ERROR_MESSAGE="APM injection is not supported on $DISTRIBUTION  < 7"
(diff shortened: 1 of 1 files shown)


install_script.sh.template
```diff
@@ -1057,6 +1057,13 @@ if [ "$OS" == "RedHat" ]; then
                 exit;
             fi
         else
+            if [ ${#apm_libraries[@]} -gt 0 ]; then
```
chouquette, line 1060: Shouldn't this outside of the innermost `if/else` (so on line 1052)?
This code is only reached when the user doesn't specify a minor vesion (ie. attempts to install the latest), while I think you'd want to error out when the user attempt to install anything through the installer on CentOS 6 and equivalent

paullegranddc: Oh yes actually. Thank you

chouquette (approved):

PR #213, opened 2024-05-23
(fleet) populate the config with the installer registry parameters

arbll:
This PR populates the config with the installer registry parameters

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -1675,20 +1675,12 @@ function update_installer_registry(){
   local registry_url="$2"
   local registry_auth="$3"
   local config_file="$4"
-  if [ -n "$registry_url" ]; then
-    printf "\033[34m\n* Adding your DD_INSTALLER_REGISTRY_URL to the $nice_flavor configuration: $config_file\n\033[0m\n"
-    if ! $sudo_cmd grep -q "^DD_INSTALLER_REGISTRY_URL:" "$config_file"; then
-      $sudo_cmd sh -c "echo 'DD_INSTALLER_REGISTRY_URL: $registry_url' >> $config_file"
+  if [ -n "$registry_url" ] || [ -n "$registry_auth" ]; then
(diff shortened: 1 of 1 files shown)


install_script.sh.template
```diff
@@ -1660,6 +1670,28 @@ function update_remote_updates(){
       $sudo_cmd sh -c "sed -i 's/^remote_updates:.*/remote_updates: $remote_updates/' $config_file"
   fi
 }
+function update_installer_registry(){
```
chouquette, line 1673: nit: could you add a newline before/after the new function?

arbll: I followed the pattern for the adjacent functions, lmk if you want newlines everywhere, happy to add

chouquette: oh, I assumed it was already the case, my bad. If not then don't bother

install_script.sh.template
```diff
@@ -1660,6 +1670,28 @@ function update_remote_updates(){
...
+  local sudo_cmd="$1"
+  local registry_url="$2"
+  local registry_auth="$3"
+  local config_file="$4"
+  if [ -n "$registry_url" ]; then
+    printf "\033[34m\n* Adding your DD_INSTALLER_REGISTRY_URL to the $nice_flavor configuration: $config_file\n\033[0m\n"
+    if ! $sudo_cmd grep -q "^DD_INSTALLER_REGISTRY_URL:" "$config_file"; then
+      $sudo_cmd sh -c "echo 'DD_INSTALLER_REGISTRY_URL: $registry_url' >> $config_file"
+    else
+      $sudo_cmd sh -c "sed -i 's/^DD_INSTALLER_REGISTRY_URL:.*/DD_INSTALLER_REGISTRY_URL: $registry_url/' $config_file"
```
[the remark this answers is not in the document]
arbll: It shouldn't with the `'` I think. There was a separate issue, it's very tricky to change those settings because they are nested YAML values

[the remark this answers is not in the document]
chouquette: Ah indeed the single quote should still be there when `sh` interprets the command, and the double is only there for the shell that spawns `sh`. LGTM then!

PR #215, opened 2024-05-24
(fleet) make DD_REMOTE_UPDATES install the installer by default

arbll:


chouquette (commented):
Looks OK but 2 other `install_installer` call sites need to also be updated AFAICS

chouquette (approved):

PR #220, opened 2024-06-05
PROF-9889: Add support for propagating DD_PROFILING_ENABLED

szegedi:
Adds support for handling `DD_PROFILING_ENABLED` in a similar vein to how `DD_IAST_ENABLED` and `DD_APPSEC_ENABLED` are propagated to `/etc/environment` in #157.

## Rationale
We want to get some customers on a private beta of profiling with SSI.

## Testing
Unit test added and confirmed to work with
```
make
python3 unit_tests/extract_functions.py
sudo ./unit_tests/test_install_script.sh
```

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -1747,6 +1747,14 @@ function manage_client_libraries_security_config(){
     $sudo_cmd sed -i -n -e '/^DD_APPSEC_SCA_ENABLED=/!p' -e "\$aDD_APPSEC_SCA_ENABLED=$sca_enable" "$etc_environment"
   fi
 }
+function manage_client_libraries_profiling_config(){
+  local sudo_cmd="$1"
+  local etc_environment="$2"
+  local profiling_enable="$3"
+  if [ -n "$profiling_enable" ]; then
+    $sudo_cmd sed -i -n -e '/^DD_PROFILING_ENABLED=/!p' -e "\$aDD_PROFILING_ENABLED=$profiling_enable" "$etc_environment"
+  fi
+}
 # "Main" configuration update
 if [ -e "$config_file" ] && [ -z "$upgrade" ]; then
   printf "\033[34m\n* Keeping old $config_file configuration file\n\033[0m\n"
@@ -1769,6 +1777,7 @@ elif [ ! "$no_agent" ]; then
 fi
 
 manage_client_libraries_security_config "$sudo_cmd" "/etc/environment" "$DD_APPSEC_ENABLED" "$DD_IAST_ENABLED" "$DD_APPSEC_SCA_ENABLED"
+manage_client_libraries_profiling_config "$sudo_cmd" "/etc/environment" "$DD_PROFILING_ENABLED"
 
 if [ ! "$no_agent" ]; then
   $sudo_cmd chown dd-agent:dd-agent "$config_file"
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -934,6 +934,7 @@ fi
 config_file_fips=$etcdirfips/datadog-fips-proxy.cfg
 system_probe_config_file=$etcdir/system-probe.yaml
 security_agent_config_file=$etcdir/security-agent.yaml
+environment_file=/etc/environment
 
 agent_major_version=AGENT_MAJOR_VERSION_PLACEHOLDER
 # shellcheck disable=SC2050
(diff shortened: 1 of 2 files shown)


install_script.sh.template
```diff
@@ -1747,6 +1748,14 @@ function manage_client_libraries_security_config(){
     $sudo_cmd sed -i -n -e '/^DD_APPSEC_SCA_ENABLED=/!p' -e "\$aDD_APPSEC_SCA_ENABLED=$sca_enable" "$etc_environment"
   fi
 }
+function manage_client_libraries_profiling_config(){
+  local sudo_cmd="$1"
+  local etc_environment="$2"
+  local profiling_enable="$3"
+  if [ -n "$profiling_enable" ]; then
+    $sudo_cmd sed -i -n -e '/^DD_PROFILING_ENABLED=/!p' -e "\$aDD_PROFILING_ENABLED=$profiling_enable" "$etc_environment"
```
r1viollet, line 1756: do we want to override existing values ?

unit_tests/test_install_script.sh
```diff
@@ -272,5 +272,17 @@ testManageClientLibrariesSecurityConfig() {
   assertEquals 0 $?
 }
 
+### Manage client libraries profiling config
+testManageClientLibrariesProfilingConfig() {
+  rm $environment_file
+  echo 'PATH="/usr/local/sbin"' > $environment_file
```
chouquette, line 278: Is this really required? I don't understand the relation with what you're testing here

install_script.sh.template
chouquette: I'd tend to say yes, currently if the user runs the install script twice, once with `DD_PROFILING_ENABLED` and once  without, they'd keep the first invocation config as far as profiling goes, while I'd expect it to be disabled following the 2nd install

szegedi: Exactly.

unit_tests/test_install_script.sh
chouquette: Could be, but I'm worried it hides potential problems if the file doesn't exist (which sounds unlikely, but not impossible)

szegedi: I can add something like
```
    if [ ! -f "$etc_environment" ]; then
      $sudo_cmd touch "$etc_environment"
    fi
```
to make sure the file is created if it doesn't exist. `manage_client_libraries_security_config()` function is also vulnerable to the file not existing, then it'd need to be touched up too. @robertpi as you wrote `manage_client_libraries_security_config()` do you have an opinion on behavior when `/etc/environment` doesn't exist?

chouquette: LGTM for this part, thanks!

install_script.sh.template
```diff
@@ -1747,6 +1748,18 @@ function manage_client_libraries_security_config(){
...
+  local sudo_cmd="$1"
+  local etc_environment="$2"
+  local profiling_enable="$3"
+  if [ -n "$profiling_enable" ]; then
+    if [ ! -s "$etc_environment" ]; then
+      $sudo_cmd echo "DD_PROFILING_ENABLED=$profiling_enable" > "$etc_environment"
+    else
+      $sudo_cmd sed -i -n -e '/^DD_PROFILING_ENABLED=/!p' -e "\$aDD_PROFILING_ENABLED=$profiling_enable" "$etc_environment"
+    fi
+  fi
```
chouquette, line 1761: AFAICS there is currently no way to disable profiling through the script

szegedi: It is true that there is no way to remove the variable from `/etc/environment` with the script, but invoking the script with `DD_PROFILING_ENABLED=false` will set that value in it, which will disable profiling for all processes that aren't launched with the env var overridden to a different value.

szegedi: Oh wait, I think I understand your [previous comment](https://github.com/DataDog/agent-linux-install-script/pull/220#discussion_r1631068976) now. You'd expect the absence of the env var in the installer invocation to explicitly disable profiling? I could even accept this to be the expected semantics, but then it'd work differently than the currently also supported `DD_APPSEC_ENABLED` and `DD_IAST_ENABLED` vars work. I think all three should be treated identically by the script.

install_script.sh.template
szegedi: Oh wait, I probably misunderstood this. You'd expect the _absence_ of the env var in the installer invocation to _explicitly_ disable profiling? I could even accept this to be the expected semantics (although it'd warrant a wider discussion), but then it'd work differently than how the currently already publicly supported DD_APPSEC_ENABLED and DD_IAST_ENABLED vars work. I think all three should be treated identically by the script.

install_script.sh.template
chouquette: Oh indeed I completely forgot about explicitly passing `DD_PROFILING_ENABLED=false`
But as you guessed I was mostly concerned by the fact that not passing the variable will not disable profiling.

I haven't looked into the other 2 variables you mentioned, but I agree it would make sense to keep a common behavior for all 3.
I suppose this can be changed in a future PR

chouquette (approved):

install_script.sh.template
r1viollet: The use case of removing the env var could still makes sense as a fallback mechanism.
Though I am not sure of the value versus switching from `true` to `false`.

r1viollet (approved):
LGTM
Thanks for allowing this!

PR #225, opened 2024-06-13
chore: dynamically write OP owned env vars

neuronull:
OPA-1564

Previously discussed here: https://github.com/DataDog/agent-linux-install-script/pull/177/files#r1571185871

My thought process:
- We already have the common prefix heuristic in OPW, so there are no changes needed there to make this work
- Because the OPW repository is private, a static list of the env vars would need to live in a public place. It could live in this repo. And certainly that is still a possibility. But again it comes back to having to remember to maintain that list, and it inevitably falling out of sync.

### DIFF
--- a/install_script_op_worker2.sh
+++ b/install_script_op_worker2.sh
@@ -164,64 +164,6 @@ if [ -n "$DD_INSTALL_ONLY" ]; then
     no_start=true
 fi
 
-# sources
-source_splunk_hec_address=
-if [ -n "$DD_OP_SOURCE_SPLUNK_HEC_ADDRESS" ]; then
-    source_splunk_hec_address="$DD_OP_SOURCE_SPLUNK_HEC_ADDRESS"
-fi
-
-source_splunk_tcp_address=
-if [ -n "$DD_OP_SOURCE_SPLUNK_TCP_ADDRESS" ]; then
-    source_splunk_tcp_address="$DD_OP_SOURCE_SPLUNK_TCP_ADDRESS"
-fi
-
-source_datadog_agent_address=
-if [ -n "$DD_OP_SOURCE_DATADOG_AGENT_ADDRESS" ]; then
-    source_datadog_agent_address="$DD_OP_SOURCE_DATADOG_AGENT_ADDRESS"
-fi
-
-source_sumo_logic_address=
-if [ -n "$DD_OP_SOURCE_SUMO_LOGIC_ADDRESS" ]; then
-    source_sumo_logic_address="$DD_OP_SOURCE_SUMO_LOGIC_ADDRESS"
-fi
-
-# source key passes
-source_splunk_tcp_key_pass=
-if [ -n "$DD_OP_SOURCE_SPLUNK_TCP_KEY_PASS" ]; then
-    source_splunk_tcp_key_pass="$DD_OP_SOURCE_SPLUNK_TCP_KEY_PASS"
-fi
-
-source_splunk_hec_key_pass=
-if [ -n "$DD_OP_SOURCE_SPLUNK_HEC_KEY_PASS" ]; then
-    source_splunk_hec_key_pass="$DD_OP_SOURCE_SPLUNK_HEC_KEY_PASS"
-fi
-
-# destinations
-dest_splunk_hec_endpoint=
-if [ -n "$DD_OP_DESTINATION_SPLUNK_HEC_ENDPOINT_URL" ]; then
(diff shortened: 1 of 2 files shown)


install_script_op_worker2.sh
```diff
@@ -524,73 +466,16 @@ else
...
-    printf "\033[34m  * Assigning DD_OP_DESTINATION_DATADOG_ARCHIVES_AWS_ACCESS_KEY_ID.\n\033[0m\n"
-    $sudo_cmd sh -c "echo DD_OP_DESTINATION_DATADOG_ARCHIVES_AWS_ACCESS_KEY_ID=$dest_datadog_archives_aws_key_id>> $env_file"
-  fi
-
-  if [ "$dest_datadog_archives_aws_secret" ]; then
-    printf "\033[34m  * Assigning DD_OP_DESTINATION_DATADOG_ARCHIVES_AWS_SECRET_ACCESS_KEY.\n\033[0m\n"
-    $sudo_cmd sh -c "echo DD_OP_DESTINATION_DATADOG_ARCHIVES_AWS_SECRET_ACCESS_KEY=$dest_datadog_archives_aws_secret>> $env_file"
-  fi
+  # All env vars that are defined and used exclusively in OPW, are prefixed with `DD_OP_`
+  for dd_op_var in ${!DD_OP_@}; do
```
spencergilbert, line 475: **❓ question:** ‏What's OPW's behavior if it encounters unknown variables/configuration here?

neuronull: My understanding is that this env file is only used in conjunction with systemd.
OPW is only checking values for the env vars it cares about/knows about.
So if there is any env var that OPW doesn't explicitly read that is written in the env file, it goes ignored.

spencergilbert: That sounds right from an "outside-OPW" angle - there's no UX concern about someone _trying_ to set a variable that isn't supported and wondering why OPW isn't working as expected?

I assume that's _fine_, but wanted to double check the product/UX side of it.

spencergilbert (commented):
Could we also add a test for this? With both the happy and sad paths?

install_script_op_worker2.sh
neuronull: I think I understand the scenario you are depicting. I could see it as a nice-to-have UX. With where we are right now, I don't think it's an issue to not be able to detect when the user is trying to export a DD_OP_ prefixed env var that we don't actually support. FWIW we don't have that functionality today without this PR. I could see it being a handy enhancement though. We could consider flagging that warning from the worker itself, and avoid having to maintain a special public list of the env vars.

neuronull:
> Could we also add a test for this? With both the happy and sad paths?

🤔  what would you consider the sad path case?

neuronull:
Also, trying to understand how the testing works.

To set an env var to the installation script, would I add them to the 

`variables`

?

https://github.com/DataDog/agent-linux-install-script/blob/ae07584373fda74eb19191460b68ecb5719609ae/.gitlab-ci.yml#L223

tangentially- should we change this now?

https://github.com/DataDog/agent-linux-install-script/blob/ae07584373fda74eb19191460b68ecb5719609ae/.gitlab-ci.yml#L224-L225

spencergilbert:
> 🤔 what would you consider the sad path case?

My mind went to "non-permitted" variables as "sad."

> Also, trying to understand how the testing works.
> 
> To set an env var to the installation script, would I add them to the
> 
> `variables`
> 
> ?
> 
> https://github.com/DataDog/agent-linux-install-script/blob/ae07584373fda74eb19191460b68ecb5719609ae/.gitlab-ci.yml#L223

I _think_ if you were up to it, I'd recommend an e2e test - but I don't know if you have any experience with it.

https://github.com/DataDog/agent-linux-install-script/tree/main/test/e2e

> tangentially- should we change this now?
> 
> https://github.com/DataDog/agent-linux-install-script/blob/ae07584373fda74eb19191460b68ecb5719609ae/.gitlab-ci.yml#L224-L225

Yikes, yes I definitely think we should 😅.

neuronull:
Forgot to mention but I did test this out manually and confirmed the env vars are properly discovered.

neuronull:
We chatted in slack and I created a followup JIRA task to track adding an e2e test.

spencergilbert (approved):

neuronull:
closing in favor of non-fork version of the changes: https://github.com/DataDog/agent-linux-install-script/pull/226

PR #230, opened 2024-06-24
PROF-9889: Fix for permission denied on empty /etc/environment

szegedi:
Fixes a bug in #220 where echo was incorrectly used with sudo to create `/etc/environment` when it doesn't exist.

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -1756,7 +1756,7 @@ function manage_client_libraries_profiling_config(){
   local profiling_enable="$3"
   if [ -n "$profiling_enable" ]; then
     if [ ! -s "$etc_environment" ]; then
-      $sudo_cmd echo "DD_PROFILING_ENABLED=$profiling_enable" > "$etc_environment"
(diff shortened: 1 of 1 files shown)


chouquette (approved):

f4usto (commented):
Thanks for addressing it @szegedi .

Would be possible to add an test case that cover this scenario?

szegedi:
@f4usto with regard to tests: we have a test that exercises this path, but minus the sudo, so it wouldn't have caught the permissions problem. It seems to me though that https://github.com/DataDog/agent-linux-install-script/blob/main/unit_tests/test_install_script.sh only works properly when run with sudo, so it could not be formulated to fail anyway as it always runs with extra permissions.

On the other hand, new onboarding tests I added in https://github.com/DataDog/system-tests/pull/2563 actually failed due to this bug on Amazon Linux VMs 'cause on Amazon Linux `/etc/environment` is a zero-length file, in contrast with Ubuntu where it has some content by default. I now have a workaround in my tests, but once we release this I'll remove the workaround, and then my system tests will exercise this.

PR #234, opened 2024-07-11
fix(debian): Only exclude pinned tracer versions from apt updates if they've been installed

BaptisteFoy:
Today on an Debian-like OS, when a tracer version is pinned, it's excluded from APT updates regardless of whether it has been installed or not. It worked fine before as it was always installed through APT, but now it can and will be installed with the installer by default: running apt-mark will fail the install script.

This PR fixes it by checking that the tracer has been installed through APT before running apt-mark. It also fixes injection tests.

### DIFF
--- a/test/localtest.sh
+++ b/test/localtest.sh
@@ -167,20 +167,20 @@ fi
 
 if [ -n "$DD_APM_INSTRUMENTATION_ENABLED" ] || [ "${SCRIPT_FLAVOR}" == "docker_injection" ]; then
   if [[ "$OS_TYPE" == "ubuntu" ]]; then
-      ls /opt/datadog-packages/datadog-apm-inject || debsums -c datadog-apm-inject
-      ls /opt/datadog-packages/datadog-apm-library-dotnet || debsums -c datadog-apm-library-dotnet
-      ls /opt/datadog-packages/datadog-apm-library-java || debsums -c datadog-apm-library-java
-      ls /opt/datadog-packages/datadog-apm-library-js || debsums -c datadog-apm-library-js
-      ls /opt/datadog-packages/datadog-apm-library-python || debsums -c datadog-apm-library-python
-      ls /opt/datadog-packages/datadog-apm-library-ruby || debsums -c datadog-apm-library-ruby
+      test -d /opt/datadog-packages/datadog-apm-inject/stable || debsums -c datadog-apm-inject
+      test -d /opt/datadog-packages/datadog-apm-library-dotnet/stable || debsums -c datadog-apm-library-dotnet
+      test -d /opt/datadog-packages/datadog-apm-library-java/stable || debsums -c datadog-apm-library-java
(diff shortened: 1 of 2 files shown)


test/localtest.sh
```diff
@@ -167,20 +167,20 @@ fi
 
 if [ -n "$DD_APM_INSTRUMENTATION_ENABLED" ] || [ "${SCRIPT_FLAVOR}" == "docker_injection" ]; then
   if [[ "$OS_TYPE" == "ubuntu" ]]; then
-      debsums -c datadog-apm-inject
-      debsums -c datadog-apm-library-dotnet
-      debsums -c datadog-apm-library-java
-      debsums -c datadog-apm-library-js
-      debsums -c datadog-apm-library-python
-      debsums -c datadog-apm-library-ruby
+      ls /opt/datadog-packages/datadog-apm-inject || debsums -c datadog-apm-inject
```
chouquette, line 170: `test -f` would be more idiomatic and explicit

test/localtest.sh
```diff
@@ -206,19 +206,19 @@ fi
...
-    rpm --verify --nomode --nouser --nogroup datadog-apm-library-js
-    rpm --verify --nomode --nouser --nogroup datadog-apm-library-python
-    rpm --verify --nomode --nouser --nogroup datadog-apm-library-ruby
-    echo "[OK] Inject libraries installed"
+      ls /opt/datadog-packages/datadog-apm-library-dotnet || rpm --verify --nomode --nouser --nogroup datadog-apm-library-dotnet
+      ls /opt/datadog-packages/datadog-apm-library-java || rpm --verify --nomode --nouser --nogroup datadog-apm-library-java
+      ls /opt/datadog-packages/datadog-apm-library-js || rpm --verify --nomode --nouser --nogroup datadog-apm-library-js
+      ls /opt/datadog-packages/datadog-apm-library-python || rpm --verify --nomode --nouser --nogroup datadog-apm-library-python
+      ls /opt/datadog-packages/datadog-apm-library-ruby || rpm --verify --nomode --nouser --nogroup datadog-apm-library-ruby
+      echo "[OK] Inject libraries installed"
```
chouquette, line 221: I see and appreciate a proud member of the 4 spaces as tabs team :saluting_face:

chouquette (approved):
LGTM, the test -f comment isn't blocking at all

maciej-chime:
This change appears to be breaking the install script for us:

Error with `v1.32.2`:
```
$ sudo service datadog-agent start

datadog-agent: unrecognized service
Exited with code exit status 1
```

Everything worked fine just before this change was published with `v1.32.1`.

I have provided more details via a Datadog support ticket just earlier.

chouquette:
Hello :wave: 

I think your issue is related to a regression introduced in datadog-agent 7.55.0 which broke our SysVinit services
See https://github.com/DataDog/datadog-agent/issues/27523#issuecomment-2228719826
I should now be fixed with agent 7.55.1

PR #243, opened 2024-07-24
fix(installer): Don't override installer install_info

BaptisteFoy:
When installing the agent with the installer, the install info is set [here](https://github.com/DataDog/datadog-agent/blob/1d08a6a9783fe271ea3813ddf9abf60244abdf2c/pkg/fleet/installer/service/datadog_agent.go#L106-L108).

Today it's overridden later on in the install script: this PR fixes it.


chouquette (approved):
LGTM!

PR #244, opened 2024-07-25
[BARX-454] Ensure the Agent is not launched when DD_INSTALL_ONLY is set

AliDatadog:
For deb/ubuntu:

apt-get install will try to start the services after installing them. One way to stop it is to set RUNLEVEL=1 but it won't work with recent distributions. The recommended way is to use a policy-rc.d script which is done in this PR. However it [will not work for ubuntu chroots](https://serverfault.com/questions/567474/how-can-i-install-packages-without-starting-their-associated-services) (known limitation).

For RPM packages: 

[These](https://github.com/DataDog/datadog-agent/blob/def18e3815c6cc67d8669408032961804821687a/omnibus/package-scripts/agent-rpm/posttrans#L3-L5) post install scripts try to install the Agent. We set `SYSTEMD_OFFLINE=1` when running them. 
Another solution would be to just stop the service after installation.

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -1375,7 +1375,13 @@ If the cause is unclear, please contact Datadog support.
 
     echo -e "  \033[33mInstalling package(s): ${packages[*]}\n\033[0m"
 
-    $sudo_cmd apt-get install -o Acquire::Retries="5" -y --force-yes "${packages[@]}" 2> >(tee /tmp/ddog_install_error_msg >&2)
+    # apt-get will automatically start the service after installation
+    # As a workaround, we can use policy.d to prevent this behavior
+    POLICYRCD=/tmp/policy-do-not-start-service-rc.d
+    echo exit 101 > "${POLICYRCD}"
+    $sudo_cmd chmod +x "${POLICYRCD}"
+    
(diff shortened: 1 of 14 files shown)


chouquette:
A quick note before reviewing, I suppose you mean `not launched` in the PR title :grin:

install_script_op_worker2.sh
spencergilbert, line 1: **❓ question:** ‏ why isn't the RPM package fix implemented in this (and the OPWv1) scripts?

install_script.sh.template
```diff
@@ -1922,7 +1931,9 @@ for current_service in "${services[@]}"; do
     You will have to do it manually using the following command:
 
     $start_instructions\033[0m\n\n"
-
+    # It is possible that services were started during the installation phase, for example on CentOS7 (systemd < v237)
+    # In that case, we should stop them
+    $stop_instructions || true
```
spencergilbert, line 1936: **💭 thought:** ‏ should we check if it's running and then try and stop it if it is, rather than always trying to stop it?

AliDatadog: IMO maybe we should check if they are running before running any `zypper install` command. This way, it would not stop a running agent service

install_script_op_worker2.sh
AliDatadog: Forgot to remove the fix here. IMO it should be in a separate PR.

spencergilbert (approved):

PR #247, opened 2024-07-29
(fleet) enable remote updates on existing installs

arbll:
The current code only modifies datadog.yaml if the file has not been created yet. For the installer settings we need to be able to migrate and as such we need to override the agent settings.


chouquette (approved):
LGTM
Although while reviewing it seems that `DD_INSTALLER_REGISTRY` and `DD_INSTALLER_REGISTRY_AUTH` can't be updated using the install script. 
Is this expected?

PR #251, opened 2024-08-05
Pass flavor to tests

alopezz:
At some point, the passing of the flavor to the bash-based tests was lost and we're installing the same datadog-agent in all testing jobs.

This fixes that by passing the necessary environment variable. It also separates Agent 6 jobs into a separate set of jobs, because:
- It doesn't make sense to test Agent 6 on "alternative flavors", the test script will just simply pass without doing anything, which is a bit unnecessary (why spin up a bunch of runners just to do nothing).
- As we're no longer releasing Agent 6 from datadog-agent's main, this separation can be a first step towards adding some rule to not trigger Agent 6 tests when not applicable.

The regression was probably introduced in https://github.com/DataDog/agent-linux-install-script/pull/121.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -70,6 +70,7 @@ unit_tests:
   dependencies: ["generate-scripts"]
   variables:
     SCRIPT: "install_script_agent${MAJOR_VERSION}.sh"
+    DD_AGENT_FLAVOR: $FLAVOR
   script:
     - ./test/localtest.sh
   # retry the job in case of failure, we might have network issues downloading the image
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -128,6 +128,8 @@ test_centos6:
 
 test:
   extends: .test
+  variables:
+    MAJOR_VERSION: 7
   parallel:
     # NOTE: to keep the matrix reasonably sized, we don't test everything everywhere
     # for example, it's ok to test getting specific flavor/minor version only on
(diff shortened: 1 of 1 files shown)


.gitlab-ci.yml
```diff
@@ -131,35 +135,62 @@ test:
...
       - artifacts/
 
+test_agent6:
+  extends: .test
+  parallel:
+    # NOTE: Agent 6 only exists for the datadog-agent "flavor".
+    # Since we've dropped Agent 6 from the regular release cadence, it makes sense to separate these
+    matrix:
+      - IMAGE: centos:centos7
+        MAJOR_VERSION: 6
```
chouquette, line 174: Same question about `MAJOR_VERSION` here

.gitlab-ci.yml
```diff
@@ -131,35 +135,62 @@ test:
     matrix:
       - IMAGE: centos:centos7
         FLAVOR: [datadog-agent, datadog-dogstatsd, datadog-iot-agent]
-        MAJOR_VERSION: [6, 7]
+        MAJOR_VERSION: 7
```
chouquette, line 138: Would it be possible to move MAJOR_VERSION out of the matrix to just have the images in there?

chouquette (approved):

.gitlab-ci.yml
alopezz: Indeed, that should be possible, good catch.

PR #256, opened 2024-08-16
Add unconditional retries for bash-based tests that run yum

alopezz:
Mitigation for #incident-29487.

We've seen for a while occasional flakes on the shell-based tests on yum-based systems, such as https://gitlab.ddbuild.io/DataDog/agent-linux-install-script/-/jobs/598453648, which fail with a `Exiting on user cancel` message.

Since we're unable to find the source of the issue but the jobs consistently succeed on retry, we're adding some unconditional retries for the affected jobs as a workaround.


alopezz:
Turns out we already have this configured on the `.test` template, so this wouldn't have any effect, closing.

PR #257, opened 2024-08-21
Add retry mechanism to fix flaky test

AliDatadog:
This PR adds a retry mechanism to `assertUninstall` in the first commit. It should hopefully fix https://gitlab.ddbuild.io/DataDog/agent-linux-install-script/-/jobs/611768206. This test is flaky only in Redhat 8 because the `process-agent.pid` file is still there. I suppose it could be due to a race condition.

In the second commit it removes the `t.Helper()` function calls.
> Helper marks the calling function as a test helper function. When printing file and line information, that function will be skipped.

IMO `Helper()` should only be used in rare cases.


Pythyu (approved):

test/e2e/common.go
```diff
@@ -251,28 +247,27 @@ func (s *linuxInstallerTestSuite) uninstall() {
...
-	}
+	assert.EventuallyWithT(t, func(c *assert.CollectT) {
+		_, err := vm.Execute("id dd-agent")
+		assert.NoError(c, err, "user datadog-agent not present after remove")
+		assertFileExists(c, vm, fmt.Sprintf("/etc/%s/%s", s.baseName, s.configFile))
+		if flavor == "datadog-agent" {
+			// The custom file should still be here. All other files, including the extra integration, should be removed
+			assertFileExists(c, vm, fmt.Sprintf("%s/site-packages/testfile", s.getLatestEmbeddedPythonPath("datadog-agent")))
+			files := strings.Split(strings.TrimSuffix(vm.MustExecute("find /opt/datadog-agent -type f"), "\n"), "\n")
+			assert.Len(c, files, 1, fmt.Sprintf("/opt/datadog-agent present after remove, found %v", files))
```
pducolin, line 261: **💬 suggestion**
```suggestion
			assert.Len(c, files, 1, fmt.Sprintf("/opt/datadog-agent present after remove , found %v, expected only %s, files, fmt.Sprintf("%s/site-packages/testfile", s.getLatestEmbeddedPythonPath("datadog-agent"))"))
```
You might store `fmt.Sprintf("%s/site-packages/testfile", s.getLatestEmbeddedPythonPath("datadog-agent"))"` into a variable as it is used above too

pducolin (approved):

pducolin:
About `t.Helper` it is helpful to know which test called the helper functions, as they are called multiple times. Not having it will reduce readability of testing logs, as they will show up with the helper lines rather than the caller lines

test/e2e/common.go
AliDatadog: Thank you ! I applied it

PR #266, opened 2024-09-09
Fix Red Hat detection

chouquette:
It is spelled as 2 words in the files we inspect to determine the distribution we're running on.

Fixing the search pattern changes the value assigned to DISTRIBUTION so all existing checks have been updated.

OS value was also update for consistency, so we don't have to wonder if we need to spell Red Hat as one or two words

Before this change, `DISTRIBUTION` is set to `Linux` on Red Hat, which circumvents the check to ensure we pin the version to `7.51.x` on RHEL 6


chouquette:
I forgot to mention that this is linked to https://datadoghq.atlassian.net/browse/AGENT-12202

spencergilbert (approved):

PR #269, opened 2024-09-16
Allow apm install without updating the agent

raphaelgavache:
Installation of APM libraries requires re-running the install script.
This PR allows installing APM libraries without updating the agent or an involved step to avoid an update (fetch the already installed version and pin Major + Minor parameters)


chouquette (approved):
LGTM although I must say I don't understand why it was made an unsupported combination :thinking:

PR #273, opened 2024-09-20
don't attempt to install apt-transport-https with APT >= 1.5.0

chouquette:
This dependency appears to cause issues on some systems where `systemd` is held. 

Since this dependency is only a transitive dependency starting from apt 1.5, we don't need it then


chouquette:
ubuntu 22.04: 
image

ubuntu 14:
image

f4usto (approved):

PR #275, opened 2024-09-23
Add back support for APM SSI centos6

raphaelgavache:
Add back support for APM SSI centos6


chouquette:
Could you provide a more descriptive commit message? It will be used when generating the release changelog, and it isn't super explicit as such :sweat_smile:

chouquette (approved):

PR #281, opened 2024-10-01
chore: [BARX-576] Don't run test_agent6 jobs from datadog-agent pipelines

spencergilbert:
This also removes the duplicated `retry` option set on the .test job.

[Example of failure](https://gitlab.ddbuild.io/DataDog/agent-linux-install-script/-/pipelines/45527302) as I remove Agent 6 jobs from `datadog-agent`


sabrina-datadog (approved):
LGTM

PR #282, opened 2024-10-07
(fleet) add network healthchecks to the script

arbll:
This PR adds a couple network healthcheck to URLs we require to work for some features (APM instrumentation today). This should help diagnose if those urls are accessible.


raphaelgavache (approved):

BaptisteFoy (approved):

chouquette (approved):
Couldn't we achieve the same effect by issuing a `HEAD` instead of fetching the entire script & manifests?

LGTM otherwise

PR #285, opened 2024-10-09
(fleet) fix network telemetry

arbll:
This PR fixes the forwarding of the network metadata. We're hitting some issues with `.` that are interpreted in domain names. Replacing those by `_` to fix the issue.


arbll:
That way I can avoid a separate PR (https://github.com/DataDog/dd-go/pull/153429), those fields don't need explicit mapping to make it to our telemetry

chouquette (approved):

PR #292, opened 2024-10-10
fix(test): Use specific version of centos7 image without usage of http://mirrorlist.centos.org/

chouetz:
Change the image used for `centos7` everywhere. This new image do not rely anymore on `http://mirrorlist.centos.org/`
Created this change to deal with [errors](https://gitlab.ddbuild.io/DataDog/agent-linux-install-script/-/jobs/667894455#L84) detected on the `main` branch of `datadog-agent`
Pipeline is green


spencergilbert (approved):
Where's the centos7 image being maintained?

chouquette:
> Where's the centos7 image being maintained?

Here: https://github.com/DataDog/images/blob/master/base/centos7/Dockerfile

PR #299, opened 2024-11-07
Tests to ensure that deb/rpm are not installed

raphaelgavache:
The test suite `install_script_docker_injection.sh` was failing as DD_APM_INSTRUMENTATION_ENABLED="docker" is expected to be propagated, yet the template just sets it a local var. Making datadog-installer miss the env

It's not a scenario expected in customer environments as setting `DD_APM_INSTRUMENTATION_ENABLED="docker"` is always documented and we document the general `install_script.sh`

I'll discard `install_script_docker_injection.sh` after next release as its not supported for the GA. Making sure first with another release that the env is properly set

### DIFF
--- a/test/localtest.sh
+++ b/test/localtest.sh
@@ -17,6 +17,8 @@ fi
 
 cp "$SCRIPT" /tmp/script.sh
 if [ "$DD_APM_INSTRUMENTATION_ENABLED" == "all" ] || [ "$DD_APM_INSTRUMENTATION_ENABLED" == "docker" ] || [ "$SCRIPT_FLAVOR" == "docker_injection" ]; then
+    # fake presence of docker for the installer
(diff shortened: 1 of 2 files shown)


test/localtest.sh
```diff
@@ -17,6 +17,8 @@ fi
 
 cp "$SCRIPT" /tmp/script.sh
 if [ "$DD_APM_INSTRUMENTATION_ENABLED" == "all" ] || [ "$DD_APM_INSTRUMENTATION_ENABLED" == "docker" ] || [ "$SCRIPT_FLAVOR" == "docker_injection" ]; then
+    # fake presence of docker for the installer
```
raphaelgavache, line 20: docker is searched in path

spencergilbert (approved):

PR #327, opened 2025-04-17
[AGENTRUN-224] Modify DD_FIPS_MODE to install the datadog-fips-agent instead of fips-proxy

jeremy-hanna:
Here is a recording of the script removing the existing fips-proxy and agent install then installing the datadog-fips-agent successfully: https://drive.google.com/file/d/13hVPU5Gfmsa2g4vIAc0k-5pzi1o0uYy1/view?usp=sharing

* Remove fips-proxy from install script
* Remove existing non-fips datadog-agent and datadog-fips-proxy packages if present when trying to install the datadog-fips-agent

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -283,9 +283,6 @@ function getMapData() {
              "datadog-dogstatsd")
                 DATA="Datadog Dogstatsd"
                 ;;
-             "datadog-fips-proxy")
-                DATA="Datadog FIPS Proxy"
-                ;;
              "datadog-fips-agent")
                 DATA="Datadog FIPS Agent"
                 ;;
@@ -646,11 +643,6 @@ if [ -n "$DD_UPGRADE" ]; then
   upgrade=$DD_UPGRADE
 fi
 
-fips_mode=
-if [ -n "$DD_FIPS_MODE" ]; then
-  fips_mode=$DD_FIPS_MODE
-fi
-
 error_tracking_standalone=
 if [ -n "$DD_APM_ERROR_TRACKING_STANDALONE" ]; then
   error_tracking_standalone=$DD_APM_ERROR_TRACKING_STANDALONE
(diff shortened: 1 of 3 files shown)


install_script.sh.template
```diff
@@ -590,6 +587,49 @@ function install_installer() {
     return "$exit_status"
 }
 
+function remove_existing_packages_for_fips_flavor() {
+  local sudo_cmd="$1"
+  local os="$2"
+  local distribution="$3"
+  local packages_to_uninstall="$4"
```
celenechang, line 594: are these args needed?

jeremy-hanna: good catch, removed in https://github.com/DataDog/agent-linux-install-script/pull/327/commits/45b0dd178ffb9075f5487a85ac37e7bebb96681f

install_script.sh.template
```diff
@@ -590,6 +590,38 @@ function install_installer() {
...
+        return 0
+    fi
+
+    if [ "$exit_status" -ne 0 ]; then
+        return 0
+    fi
+
+    echo -e "\033[34m\n* Removing the datadog-fips-proxy and existing datadog-agent packages\n\033[0m"
+    if [ "$os" == "Debian" ]; then
+        $sudo_cmd apt-get remove -y --force-yes "datadog-fips-proxy" "datadog-agent" || return $?
```
celenechang, line 615: will an error message surface if it returns a nonzero exit status?

jeremy-hanna: Hmm we're not redirecting stderr with these commands so it should be visible in the script? I can add out own error message here, we probably would not want to silently continue on a nonzero exit code (tbh I don't know what could cause a `apt-get|yum remove -y` could fail with a nonzero)

jeremy-hanna: added in https://github.com/DataDog/agent-linux-install-script/pull/327/commits/ae3c26717fc1bbfe80125409fa149455e66adc66

celenechang: thanks!

celenechang (approved):

PR #336, opened 2025-05-27
(fleet) trigger installer trace from exit trap

arbll:
This PR triggers the installer trace from an EXIT trap, ensuring it's always called.


install_script.sh.template
```diff
@@ -378,8 +404,8 @@ function report_installer_telemetry() {
         "origin": "linux-install-script",
         "host": {
             "hostname": "$(json_escape "$(uname -n)")",
-            "os": "$(json_escape "${os}")",
-            "distribution": "$(json_escape "${distribution}")",
+            "os": "$(json_escape "$(uname -o)")",
+            "distribution": "$(json_escape "$(lsb_release -ds)")",
```
celenechang, line 408: will `lsb_release` work on all OSs?

PR #337, opened 2025-05-27
upgrade(fleet): Use installer install script to install APM SSI

BaptisteFoy:
There is a new `install-ssi.sh` flavor of the installer install script available to install only SSI. This PR uses it ; so we can deprecate the Datadog Installer deb & rpm packages

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -9,7 +9,7 @@ set -e
 
 DEPRECATION_MESSAGE_PLACEHOLDER
 
-install_script_version=1.39.0.post
+install_script_version=1.40.0.post
 logfile="ddagent-install.log"
 support_email=support@datadoghq.com
 variant=install_scriptINSTALL_INFO_VERSION_PLACEHOLDER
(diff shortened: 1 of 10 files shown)


arbll (approved):

install_script.sh.template
```diff
@@ -777,22 +738,6 @@ if [ -n "$DD_AGENT_FLAVOR" ]; then
     agent_flavor=$DD_AGENT_FLAVOR #Eg: datadog-iot-agent
 fi
 
-datadog_installer=
-if [ -n "$DD_INSTALLER" ]; then
```
betterengineering, line 781: I'm not sure how this is supposed to work with the refactors, but this is likely why we needed this change:  https://github.com/DataDog/auto_inject/pull/774

install_script.sh.template
```diff
@@ -577,39 +553,6 @@ function filter_packages_installed_by_installer() {
...
-    local remote_updates="$5"
-    local apm_instrumentation_enabled="$6"
-    local pkg_array_name="$7"
-
-    local trace_id
-    local start_time
-    local exit_status
-
-    # The installer is currently only being rolled out to APM instrumentation users
-    if [ -z "$apm_instrumentation_enabled" ] && [ -z "$datadog_installer" ] && [ -z "$remote_updates" ]; then
```
betterengineering, line 595: I see. This line also allowed us to install the installer without installing APM by setting `DD_INSTALLER=true`

PR #341, opened 2025-06-12
Move old script to its own Gitlab stage - deprecated

tbavelier:
### What does this PR do ?

* Adds a new Gitlab stage `deploy_deprecated` for the `install_script.sh` that is rarely updated since it's deprecated.

### Motivation

* Clearly separates the old script publish step to avoid triggering it by mistake as all other jobs part of `deploy` step are manually triggered

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,6 +5,7 @@ stages:
   - e2e
   - e2e_test_upload
   - deploy
+  - deploy_deprecated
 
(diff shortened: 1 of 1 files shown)


tbavelier:
Closing in favour of https://github.com/DataDog/agent-linux-install-script/pull/345 (same PR, but had manually cancelled the pipeline on that one so checks couldn't run)

PR #342, opened 2025-06-19
fix(tests): Remove unneeded broken test

BaptisteFoy:
This test breaks because:
- Removing the installer deb calls `datadog-installer purge`
- Starting agent 7.67.0; datadog-installer is symlinked to the installer in the agent
- That installer in the agent calls preRemove on the agent package
- This stops the agent

As installing the deb standalone will soon not be supported anymore; we'll simply remove the failing test.

### DIFF
--- a/test/e2e/install_installer_test.go
+++ b/test/e2e/install_installer_test.go
@@ -146,47 +146,6 @@ func (s *installUpdaterTestSuite) purge() {
 	}
 }
 
-func (s *installUpdaterTestSuite) assertInstallerInstalled() {
-	t := s.T()
-	vm := s.Env().RemoteHost
-
-	t.Log("Assert installer is installed")
-	assertFileExists(t, vm, "/opt/datadog-packages/datadog-installer/stable/bin/installer/installer")
-	assertFileExists(t, vm, "/opt/datadog-installer/bin/installer/installer")
-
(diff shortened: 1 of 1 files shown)


test/e2e/install_installer_test.go
```diff
@@ -36,24 +36,6 @@ func TestInstallUpdaterSuite(t *testing.T) {
...
-func (s *installUpdaterTestSuite) TestInstallUpdater() {
-	t := s.T()
-	vm := s.Env().RemoteHost
-	cmd := fmt.Sprintf("DD_INSTALLER=true DD_API_KEY=%s DD_SITE=\"datadoghq.com\" bash -c \"$(cat scripts/install_script_agent7.sh)\"", apiKey)
-	output := vm.MustExecute(cmd)
-	t.Log(output)
-	defer s.purge()
-
-	s.assertInstallScript(true)
-	s.assertInstallerInstalled()
```
tbavelier, line 48: I believe `assertInstallerInstalled` is solely used in this removed function, could we remove it too ? Same for `assertUninstallInstaller`, `uninstallInstaller`

tbavelier (changes requested):

test/e2e/install_installer_test.go
BaptisteFoy: Ah yes good catch, thanks

PR #365, opened 2025-08-26
Enable logs_config.auto_multi_line_detection for process logs

vitkyrka:
Add auto_multi_line_detection: true to the logs_config section when enabling process log collection. This enables automatic detection of multi-line log entries for process logs.

Updated tests to verify the new configuration is properly set.

https://datadoghq.atlassian.net/browse/DSCVR-184

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -1800,6 +1800,7 @@ logs_enabled: true
 
 logs_config:
   process_exclude_agent: true
+  auto_multi_line_detection: true
 
 process_config:
   process_collection:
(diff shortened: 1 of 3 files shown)


tbavelier (changes requested):
At this stage, this would be equivalent to changing https://github.com/DataDog/datadog-agent/blob/4b6da502473f130c936b0444408882d008faa2cd/pkg/config/setup/config.go#L1731 to `true`, as it will apply for all logs, not only the Agent own logs.
TBD

PR #436, opened 2026-07-02
Revert "Make sure we keep the testing params when extra params does not override minor version"

KevinFairise2:
Reverts DataDog/agent-linux-install-script#430


oleroydatadog (approved):

quentinus95:
@oleroydatadog to summary: the test is succeeding but the nightly pipelines are not testing the agent artifact built in the pipeline, only testing the latest stable. You should plan to update and fix the tests, including re-applying this change to ensure you properly test the install script in this scenario against the proper release candidate. If the `datadog-fips-proxy` related tests are not important to your team, a trade-off could be to disable or simplify those tests while not impacting the rest of the suite.

PR #437, opened 2026-07-20
Add DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT env var

julesmcrt:
## Summary
- Adds a new `DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT` env var to set `private_action_runner.api_key_only_enrollment` in `datadog.yaml` at install time, following the same pattern as the existing `DD_PRIVATE_ACTION_RUNNER_ACTIONS_ALLOWLIST` env var.

## Test plan
- [x] Added unit tests covering PAR enabled with and without `api_key_only_enrollment` set (`testParEnabledWithApiKeyOnlyEnrollment`, `testParEnabledWithoutApiKeyOnlyEnrollment`)

### DIFF
--- a/install_script.sh.template
+++ b/install_script.sh.template
@@ -976,6 +976,11 @@ if [ -n "$DD_PRIVATE_ACTION_RUNNER_ACTIONS_ALLOWLIST" ]; then
   par_actions_allowlist=$DD_PRIVATE_ACTION_RUNNER_ACTIONS_ALLOWLIST
 fi
 
+par_api_key_only_enrollment=
+if [ -n "$DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT" ]; then
+  par_api_key_only_enrollment=$DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT
+fi
+
 ##
 # INSTALL SCRIPT CONFIGURATION OPTIONS
 # Technical options to test with non-production values for signature keys, packages or reporting telemetry.
@@ -2211,6 +2216,7 @@ function update_par(){
   local config_file="$2"
   local par_enabled="$3"
   local par_actions_allowlist="$4"
+  local par_api_key_only_enrollment="$5"
 
   if [ "$par_enabled" != "true" ]; then
     return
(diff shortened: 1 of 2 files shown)


merchristK (approved):
lgtm. thks

install_script.sh.template
```diff
@@ -976,6 +976,11 @@ if [ -n "$DD_PRIVATE_ACTION_RUNNER_ACTIONS_ALLOWLIST" ]; then
   par_actions_allowlist=$DD_PRIVATE_ACTION_RUNNER_ACTIONS_ALLOWLIST
 fi
 
+par_api_key_only_enrollment=
+if [ -n "$DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT" ]; then
+  par_api_key_only_enrollment=$DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT
+fi
```
Stanislas167, line 982: If only 'true' and 'false' are available, I prefer
```suggestion
par_api_key_only_enrollment=
if [ -n "$DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT" ]; then
  par_api_key_only_enrollment=true
fi
```

at least to enforce the case/syntax.
Because currently I think the user can put whatever he wants in the config.

julesmcrt: I see what you mean, but hardcoding the value here would mean that `DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT=false` actually makes the install script enable the feature in the `datadog.yaml`!

We would need something like:
```
if [ -n "$par_api_key_only_enrollment" ] && [ "$par_api_key_only_enrollment" != "true" ] && [ "$par_api_key_only_enrollment" != "false" ]; then
  printf "\033[31m\nError: DD_PRIVATE_ACTION_RUNNER_API_KEY_ONLY_ENROLLMENT must be true or false\n\033[0m\n"
  return
fi
```
which becomes way too complicated...

And the current syntax is coherent with the other ENV VARs for PAR above

Stanislas167 (approved):

PR #446, opened 2026-08-21
Revert "Revert "Make sure we keep the testing params when extra params does not override minor version""

ofek:
Reverts DataDog/agent-linux-install-script#436

The issue is being fixed in https://github.com/DataDog/datadog-agent/pull/54908.


ofek:
/trigger-ci --variable RUN_ALL_BUILDS=true --variable RUN_KITCHEN_TESTS=true --variable RUN_E2E_TESTS=on --variable RUN_UNIT_TESTS=on --variable RUN_KMT_TESTS=on --variable DEPLOY_AGENT=true

dustmop (approved):

